PR #2, opened 2016-09-14
Fix typo in README.md

bedeoverend:
Example of retaining only selected criteria should read `hooks.pluckQuery` rather than `hooks.removeQuery`


eddyystop:
Thanks for catching that.

PR #12, opened 2016-10-19
Fixed tests which had failed in Node 4

eddyystop:
The code in src/ had no issues running on Node 4, but the tests tested
features which Node 4 doesn't support and the test code itself
failed on Node 4.
- Turned parseFunc into a wrapper over a new parseSignature.
The tests are now against parseSignature.
- Unfortunately had to inline repo get-parameter-names to make a small
change so the tests run on Node 4.


ekryski:
Looks good. Maybe we should create an issue or add a note to the code to remove it when we drop support for Node v4.

ekryski:
:shipit:

eddyystop:
Note already in code. Will create an issue also.

PR #24, opened 2016-10-27
Removed overly complex promisify function wrappers

eddyystop:
Function signatures were parsed to get the #params in a function
and to see if the func was likely a callback func. There were
some manageable issues with signatures whose params had default
values. However the parser used does not work for params using
deconstruction, and those are common. We would have to switch to
a parser that parses signatures into an Abstract SyntaxTree to
solve this problem. That is really going too far for the small
benefits any parsing provides.

- Removed all the promisify func except fnPromisfyCallback
- Renamed fnPromisifyCall to the more descriptive callbackToPromise
- Removed tests related to promisfy funcs
- Updated test for callbackToPromise
- Updated README.
- Closed issue reminding us to remove hacks to get around Node 4.


ekryski:
Looks good @eddyystop! I think this is a good call. Simpler = better.

Likely will need to be a major release I guess.

:shipit:

PR #25, opened 2016-10-27
Added promiseToCallback, perhaps more rugged than Feathers' code

eddyystop:
The callback is invoked outside the scope of the Promise.
The Promise's .catch() chain is not invoked if the callback throws.
This code is used within feathers-service-verify-reset to resolve
eddyystop/feathers-service-verify-reset#12


daffl:
👍

PR #26, opened 2016-10-28
Remove dependency on feathers-authentication

daffl:
This repository should not depend on feathers-authentication. Since the one hook that does will be deprecated anyway we should get this into the 2.0 release. Necessary for https://github.com/feathersjs/feathers-hooks/pull/120.


eddyystop:
Sure. I'll remove it from README and GitBook.

daffl:
Done in https://github.com/feathersjs/feathers-docs/commit/4b5ceac5cf100cb6f6fc486dbfcfea2377db7b45

PR #27, opened 2016-10-29
Update to latest plugin infrastructure

### LINKED ISSUE #13: Standardize scripts
Convert to these where possible https://github.com/feathersjs/generator-feathers-plugin/tree/master/app/templates

daffl:
This updates the release and tests scripts to the current infrastructure for other plugins and uses ES `import` in the tests.

Closes #13

/cc @eddyystop


eddyystop:
OK

ekryski:
👍 . Looks good to me. I'd say :shipit: but @eddyystop you make the call.

eddyystop:
Ok with me. I'll be a while before I can install rail though.

ekryski:
We don't need to rush this. I'll look into this, this week for sure. Even if it means quickly writing our own module. It's really not hard. @corymsmith and I both have written scripts using github and git to commit things.

ekryski:
I'm drawing a hard line on that. I don't want people to need to install ruby just to get this working. That will also mean anyone on windows has to do feathers dev in a VM.

@eddyystop let's merge this but just comment out the changelog generation right now. We'll run through all the repos and re-gen changelogs once I get a better node one sorted.

eddyystop:
I'm OK with merging.

PR #33, opened 2016-11-04
Fix error in populate hook description example

danieledler:
### Summary

* Mixed 'target' and 'options.field' in description example.
* Mention that the hook is ignored if ‘options.field’ doesn’t exist on
result
* Add test that checks above


ekryski:
Awesome thank you!

ekryski:
released in v2.0.1

PR #34, opened 2016-11-04
Fix bug in populate, sending wrong params

danieledler:
### Summary

Fix bug using empty params in populate hook, causing for example a remove after hook on populated item to be ignored even if called from non-internal provider.

### DIFF
--- a/src/bundled.js
+++ b/src/bundled.js
@@ -414,7 +414,7 @@ export function populate (target, options) {
         item = item.toJSON(options);
       }
       // Remove any query from params as it's not related
-      const params = Object.assign({}, params, { query: undefined });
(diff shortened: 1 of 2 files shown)


src/bundled.js
```diff
@@ -414,7 +414,7 @@ export function populate (target, options) {
         item = item.toJSON(options);
       }
       // Remove any query from params as it's not related
-      const params = Object.assign({}, params, { query: undefined });
+      const params = Object.assign({}, hook.params, { query: undefined });
```
ekryski: This is totally correct. Thanks for catching that.

ekryski:
PR looks good. Thanks @danieledler! ❤️  Nice catch 😄

ekryski:
released in v2.0.1

PR #40, opened 2016-11-08
Fix softDelete error on 'find'

jojobyte:
### Summary
Adding `softDelete()` to a before hook for find is throwing `Cannot read property 'hasOwnProperty' of undefined` which appears to be caused by trying to modify the `hook.data` param instead of `hook.params.query`


eddyystop:
Thanks for the catch!

Issue #44, opened 2016-11-22
softDelete undefined.patch issue


KidkArolis:
I just noticed that in the published npm code, the softDelete is compiled from

```js
return this.patch(hook.id, hook.data, hook.params).then(data => {
```

to

```js
return undefined.patch(hook.id, hook.data, hook.params).then(function (data) {
```

Don't think that's correct. I'll dig around more tomorrow if I get a minute.

eddyystop:
Need to add a proper test.

eddyystop:
I'm writing unit tests. `get` with the above changes returns the soft deleted item.

**Minus**: subsequent code has to check if item is deleted. **Plus**: A softDeleted item can be read and 'undeleted' without hacking with `hooks.iff()`.

I'm inclined to leave it as is. What say you @daffl ?

eddyystop:
softDelete rewritten with tests in v2.0.2

PR #45, opened 2016-11-23
Fix #30 (allow all hooks) and #44 (arrow function can't reference this)

### LINKED ISSUE #30: Allow softDelete on all methods
Not just remove & find.

KidkArolis:
Will try to add some tests as well.

### DIFF
--- a/src/new.js
+++ b/src/new.js
@@ -16,10 +16,10 @@ import { setByDot, checkContext } from './utils';
  *   find: [ softDelete() ] // ignore deleted items
  * };
  */
-export const softDelete = (field) => (hook) => {
-  checkContext(hook, 'before', ['remove', 'find'], 'softDelete');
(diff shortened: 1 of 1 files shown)


daffl:
Thank you, that definitely makes sense. Looks like that one doesn't have any tests which is probably why it slipped through.

daffl:
Thank you, that definitely makes sense. Looks like that one doesn't have any tests which is probably why it slipped through.

eddyystop:
@KidkArolis Thanks for the catch! Would you mind changing the `!=` to `!==` so the lint passes? Otherwise I can do it later today.

The other hooks are tested with fakes which weren't up to handling this hook. A test would be welcome or i can make one in a day or so.

KidkArolis:
fix linting

PR #46, opened 2016-11-23
fix setUpdatedAt and setCreatedAt

alerosa:
The docs for `setCreatedAt` and setUpdatedAt` says that they "Set the fields to the current date-time" but they set the date when the feathers application bootstraps.

That's because the `new Date` function is called during the bootstrap and not every time the hook is triggered. To solve that, I've changed the `setField` method so that now it can accept a function as second argument and it will call it when the hook is actually triggered.


eddyystop:
Thanks very much for the catch!

PR #51, opened 2016-11-28
Rewrote softDelete to properly handle all methods

eddyystop:
- service.before({ all: hooks.softDelete() }) will handle all methods
- get, plus update,patch,remove with an id, all perform a preliminary get
to determine if the item is already soft deleted. They throw if it is.
- unit tests written


daffl:
:shipit:

PR #52, opened 2016-11-28
Added support for multiple hooks in iff()

eddyystop:
- exporting doHooks which runs multi hooks using
feathers-hooks/lib/commons.processHooks
- changed iff() to call doHooks
- wrote tests for doHooks
- changed iff tests as doHooks always returns a Promise
- change isNot, isProvider tests for the same reason


daffl:
:shipit:

PR #56, opened 2016-11-29
Added a hook to execute a set of hooks

eddyystop:
- create: hooks.combine(hook1, hook2, ...) // same as [hook1, hook2, ...]
- return hooks.combine(hook1, hook2, ...).call(this, currentHook)
.then(hook => { ... });
- renamed doHooks to combine
- refactored tests
- documented combine in /* */
- closes feathersjs/feathers-hooks#68

### DIFF
--- a/src/new.js
+++ b/src/new.js
@@ -77,7 +77,25 @@ export const softDelete = field => {
   };
 };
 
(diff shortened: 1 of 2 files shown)


daffl:
👍

PR #57, opened 2016-11-30
Added support for iff().else(hook1, hook2, ...)

eddyystop:
- Implemented support
- Updated tests.

### DIFF
--- a/src/new.js
+++ b/src/new.js
@@ -100,11 +100,11 @@ export const combine = (...rest) => function (hook) {
 };
 
 /**
- * Hook to conditionally execute another hook.
+ * Hook to conditionally execute one or another set of hooks.
  *
  * @param {Function|Promise|boolean} ifFcn - Predicate function(hook).
- *    Execute hookFcn if result is truesy.
- * @param {Array.function} rest - Hook functions to execute.
+ * @param {Array.function|Function} trueHooks - Hook functions to execute when ifFcn is truesy.
+ * @param {Array.function|Function} falseHooks - Hook functions to execute when ifFcn is falsey.
  * @returns {Object} resulting hook
  *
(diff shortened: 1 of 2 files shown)


daffl:
This looks good. One thing for the tests is that you can now `return` a Promise in a test in Mocha 3. So something like

```js
it('calls async hook function if aync predicate truthy', (done) => {
  const result = hooks.iff(() => new Promise(resolve => resolve(true)), hookFcnAsync)(hook);

  if (result && typeof result.then === 'function') {
   result.then(result1 => {
      assert.deepEqual(result1, hookAfter);
      assert.equal(hookFcnAsyncCalls, 1);
      assert.deepEqual(result1, hookAfter);

      done();
    });
  } else {
    assert.fail(true, false, 'promise unexpectedly not returned');
    done();
  }
});
```

Would now look like

```js
it('calls async hook function if aync predicate truthy', () => {
  const result = hooks.iff(() => new Promise(resolve => resolve(true)), hookFcnAsync)(hook);

  if (result && typeof result.then === 'function') {
    return result.then(result1 => {
      assert.deepEqual(result1, hookAfter);
      assert.equal(hookFcnAsyncCalls, 1);
      assert.deepEqual(result1, hookAfter);
    });
  } else {
    assert.fail(true, false, 'promise unexpectedly not returned');
  }
});
```

The advantage is that you won't have to call `done` all the time and it catches all errors (the previous example with `done` would just time out if there was an assertion error).

eddyystop:
Good news about returning Promises.

daffl:
:shipit:

daffl:
Actually by shipit I mean good go merge 😉 I was going to work on finishing up some smaller changes I have locally this weekend and then we can make a 2.1.0 release.

PR #59, opened 2016-12-07
Improved perf on the most common usage of getByDot, setByDot

eddyystop:
These are used a fair amount in the serialize hook.


ekryski:
:shipit:

PR #61, opened 2016-12-07
adding a some hook

ekryski:
### Summary

Adding a utility hook that takes in other functions and resolves to `true` if any of the functions return true. Resolves to `false` if all functions return falsey values. Errors thrown from a hook function are deemed as falsey and are swallowed. This is the only thing I'm not sure of. What should it do?

**EDIT**: It now rejects errors in the hook chain properly.

Usage is like so:

```js
some(hook1, hook2, hook3)
````

### Other Information

cc/ @daffl @bedeoverend

### DIFF
--- a/src/new.js
+++ b/src/new.js
@@ -103,7 +103,7 @@ export const combine = (...rest) => function (hook) {
  * Hook to conditionally execute one or another set of hooks.
  *
  * @param {Function|Promise|boolean} ifFcn - Predicate function(hook).
- * @param {Array.function|Function} trueHooks - Hook functions to execute when ifFcn is truesy.
+ * @param {Array.function|Function} trueHooks - Hook functions to execute when ifFcn is truthy.
  * @param {Array.function|Function} falseHooks - Hook functions to execute when ifFcn is falsey.
  * @returns {Object} resulting hook
  *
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by bedeoverend** [comment]
@ekryski I'd vote no on swallowing errors, at least by default. In the context of [feathersjs/feathers-permissions#4](https://github.com/feathersjs/feathers-permissions/issues/4) it might make sense to swallow, given that I'd imagine permissions would deny on failure, but as a generic hook, I'd prefer it to crash so I could handle _unexpected_ errors.

Perhaps it should swallow only errors with status codes `40x`? i.e. implying that the error is expected, but that might be getting a little too magical.

File: src/new.js [hunk 1]
```diff
@@ -178,6 +178,44 @@ export const iff = (ifFcn, ...rest) => {
...
+ *
+ * Example 2 - called within a custom hook function
+ * function (hook) {
+ *   ...
+ *   return hooks.some(hook1, hook2, ...).call(this, currentHook)
+ *     .then(hook => { ... });
+ * }
+ */
+
+export const some = (...rest) => (hook) => {
```
bedeoverend: I think the inner arrow function `(hook) => {` should be `function(hook) {` so the right context is given for `fn.call(this, ...)` later on

ekryski:
Ya I was thinking that if an error happens you'd probably want to know about it as it's unexpected behaviour and it should error out of the `some` hook chain rather than swallow it. Unless anyone has any issues with that I'm going to update it so that it doesn't treat errors as falsey and swallow them.

ekryski:
I dunno what code climate is complaining about but I can't access the repo in there to get more details. Just get a 422 error. Everything else is all good.

ekryski:
ok I'm shipping it. I need this to wrap permissions.

eddyystop:
Sorry for not getting onto this. I'm preparing for a business supper.

ekryski:
np. I think it's pretty safe.

eddyystop:
GitHub shows on the PR that Code Climate failed. You can click a link on that line to see the CC page and that eventually leads you to the issues it found.

eddyystop:
Needs tests.

ekryski:
> GitHub shows on the PR when Code Climate failed. You can click a link on that line to see the CC page and that eventually leads you to the issues it found.

I just get this. Must be a permissions thing.

screen shot 2016-12-08 at 7 03 15 pm

> Needs tests.

Has tests and they are passing 😄

PR #62, opened 2016-12-07
Added new populate, dePopulate, serialze hooks.

eddyystop:
- Docs are in github eddyystop/feathers-hooks-populate for the moment.
- Old populate hook renamed legacyPopulate.
- New populate calls legacyPopulate based on func signature for backward compatibility.
- @marshall reviewed the README and pushed a PR.
- @ekryski "took a quick gander as well and it looks pretty good." I assume he looked at the code.
- I'll copy the README to GitBook once the code is reviewed.
- Perhaps these 3 hooks ought to have their own subpage hooks/common/populate
as the docs are not short.

### DIFF
--- a/test/populate-1deep-1child.test.js
+++ b/test/populate-1deep-1child.test.js
@@ -1,0 +1,273 @@
+
+const chai = require('chai');
+const configApp = require('../test/helpers/configApp');
+const getInitDb = require('../test/helpers/getInitDb');
+const { populate } = require('../src');
+
+const assert = chai.assert;
+
+describe('populate - include 1:1', () => {
+  let hookAfter;
+  let hookAfterArray;
+
+  let app;
+  let recommendation;
+
+  beforeEach(() => {
+    app = configApp(['posts', 'recommendation']);
+    recommendation = clone(getInitDb('recommendation').store);
+
+    hookAfter = {
+      type: 'after',
+      method: 'create',
+      params: { provider: 'rest' },
+      path: 'recommendations',
+      result: recommendation['1']
+    };
+    hookAfterArray = {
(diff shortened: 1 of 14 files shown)


test/populate-1deep-1child.test.js
```diff
@@ -0,0 +1,273 @@
...
+            parentField: 'postId',
+            childField: 'id',
+            permissions: 'for posts'
+          }
+        ]
+      };
+
+      return populate({ schema, checkPermissions })(hook)
+        .then(() => { throw new Error('was not supposed to succeed'); })
+        .catch(() => {});
```
ekryski: No need to stop this PR but I usually check that an error exists as well

```
.catch(error) => expect(error).to.not.equal(undefined))
```

ekryski:
> @ekryski "took a quick gander as well and it looks pretty good." I assume he looked at the code.

I did. Didn't see anything that stood out.

> Perhaps these 3 hooks ought to have their own subpage hooks/common/populate
as the docs are not short.

Agreed.

> New populate calls legacyPopulate based on func signature for backward compatibility.

💥  nice. I added a note about adding a deprecation warning guiding people towards the new format. Even just something like _"Calling `populate(target, options)` is now deprecated and will be removed in the future. Refer to docs.feathersjs.com for more information"_

Other than that I think it looks pretty good. I'll have to pull it down and give it a try.

test/populate-1deep-1child.test.js
eddyystop: Interesting. Info, thxs.

ekryski:
I think this is probably good to :shipit:

PR #63, opened 2016-12-07
adding an every hook

ekryski:
### Summary

Adding a utility hook that takes in other functions and resolves to `true` if all of the functions return true. Resolves to `false` if one of the functions returns a falsey value. Errors thrown from a hook function (ie. hook2) break out of the chain and are rejected appropriately.

Usage is like so:

```js
every(hook1, hook2, hook3)
````

### Other Information


ekryski:
ooops. I guess I didn't get a :shipit: on this. @eddyystop?

eddyystop:
Sorry I'm at a business supper

PR #64, opened 2016-12-07
adding when alias for iff

ekryski:
### Summary

You can now use `when(predicateHook, hook1, hook2, hook3)` interchangeably with `iff`.

### DIFF
--- a/src/new.js
+++ b/src/new.js
@@ -178,6 +178,12 @@ export const iff = (ifFcn, ...rest) => {
 };
 
 /**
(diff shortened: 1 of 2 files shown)


eddyystop:
Perhaps we should doc this in GitBook/Auk?

ekryski:
Will toss in a couple docs PRs into Auk this afternoon. I have a bunch of updates for auth + permissions so I'll do them at the same time.

eddyystop:
Look good.

PR #65, opened 2016-12-07
add $client hook

eddyystop:
- It will help people feel safe and warm inside,
instead of wondering if they’re doing something that’s a hack.
- Will put docs in GitBook once this is merged.


src/new.js
```diff
@@ -225,3 +225,49 @@ export const isNot = (predicate) => {
     return result.then(result1 => !result1);
   };
 };
+
+/**
+ * Move params from client to hook.params.
```
marshallswain: $client

src/new.js
```diff
@@ -225,3 +225,49 @@ export const isNot = (predicate) => {
...
+      });
+
+      delete params.query.$client;
+    }
+
+    return hook;
+  };
+};
+
+const reservedParamProps = ['app', 'authenticated', '__authenticated',
```
marshallswain: Should sequelize and mongoose be in this list?

ekryski: yes

ekryski: Not sure. I don't think so but I'd have to look.

ekryski: `permitted` and `__permitted` have now been removed. The permissions hooks no longer use those fields based on this discussion https://github.com/feathersjs/feathers-permissions/issues/4#issuecomment-265369288.

eddyystop: Its `hook.app` not `hook.params.app`. So `app` should be removed.

test/client.test.js
```diff
@@ -0,0 +1,197 @@
...
+      });
+
+      assert.throws(() => {
+        $client('query')(hook);
+      });
+
+      assert.throws(() => {
+        $client(
+          'app', 'authenticated', '__authenticated',
+          'permitted', '__permitted', 'provider', 'query'
```
eddyystop: Thanks for that catch!

daffl:
I'm not sure about `reservedParamProps` since the hook shouldn't really know about any other plugin that could potentially have restricted properties.

We might be able to define things like `authenticated` as read-only properties so it would throw an error if you are overriding it. That won't take care of when they are not defined but I'd go as far as saying that if you put `sequelize` or something like that explicitly into the whitelist it's your own fault.

eddyystop:
@ekryski and @marshall were proponents of the whitelist so we need their input.

ekryski:
> We might be able to define things like authenticated as read-only properties so it would throw an error if you are overriding it.

Yeah that's the plan but until I get that sorted the whitelist for that piece is a good prevention measure. So we can either hold up this hook or ship it and pull that stuff out in the future.

> That won't take care of when they are not defined but I'd go as far as saying that if you put sequelize or something like that explicitly into the whitelist it's your own fault.

I agree but then it should probably be `$sequelize` and `$mongoose` or something to denote that they are special. It's still pretty easy to inadvertently overwrite them. I think we have a few cases where it is a little too easy to accidentally do the wrong thing *cough* `.remove(null)` *cough*

marshall:
you have the wrong Marshall :)

ekryski:
Poor @marshall, pretty sure that has happened a couple times on some of our issues/PRs. Sorry about that mate! 😄 

@eddyystop I think you meant @marshallswain

daffl:
It's not implicit thought right? You have to explicitly whitelist a property. So if you explicitly whitelist something potentially dangerous there isn't much we can do about it.

Oh also, we probably want a reverse of this hook for the client side (so something that puts a whitelist of params into `query.$client`).

eddyystop:
@daffl, your first comment is valid. I think another point of the whitelist was for the server to limit what the client may specify. It prevents what someone hacking from the browser console can do.

How will we package client side hooks? We'd also need a copy of them in this repo so the server can do what the client can.

Merry Christmas eve.

daffl:
Probably fairly simple like:

```js
function client(... whitelist) {
  return function(hook) {
    const client = hook.params.query.$client = {};

    whitelist.forEach(key => client[key] = hook.params[key]);
  }
}
```

eddyystop:
I add the client-side hook as a separate PR

daffl:
I still don't think that the `reservedParamProps` is necessary when people have to explicitly provide a whitelist (as I said, if someone is explicitly doing something wrong we can't help them) but if it's still controversial then :shipit:

ekryski:
ok 2 comments:

- I think we should name the hook function `client` instead of `$client`. That's me being a bit pedantic but it will keep it consistent with the rest of the hooks. It looks a bit out of place being prefixed with `$`.
- After reviewing again and thinking about it, I think I'm with @daffl on this now. We don't really need the `reservedParamProps` if someone whitelists them, themselves then it's their own fault and they are whitelisting some pretty buried props. I'll work to quickly prefix them with `__` to have a bit more obscurity.

After that is done :shipit:

ekryski:
Looks good. :shipit: @eddyystop!

PR #66, opened 2016-12-07
adding unless hook

ekryski:
### Summary

Adding a utility hook that takes in a predicate function and a set of hooks. If the predicate function returns a falsey value the hooks are executed, otherwise they are not. Synonymous with `iff(isNot(fn), hook1, hook2, hook3)` except that the predicate does not need to be a function.

Usage is like so:

```js
unless(predicate, hook2, hook3)
````

### Other Information


eddyystop:
Look good. 

Should have tests though, along with the other recent hooks.

ekryski:
They all have tests and are all passing 😄 . Ok shipping!

PR #68, opened 2016-12-08
Added validateSchema hook to validate JSON objects

eddyystop:
- Uses ajv validator repo which Best Buy API playground uses
- ajv is injected into hook so its not pulled into feathers-hooks-common
- This hook differs enough not to have to attribute to BB's hook
- Handles sync ajv validation. Async results from custom plugins to ajv.
- Throws with an array of err messages in err.errors
- Overriding error formatting func could create e.g. { name: msg } type
- The above opens possibility to use this for some UI validation.


daffl:
:shipit:

ekryski:
Totally :shipit: on this one! This is sweet! 🍭

Issue #37, opened 2016-11-07
utility hook to trim data


petermikitsh:
It would be nice to have a hook that deeply inspects `hook.data` and removes whitespace. Node already has the `String.trim` function, so this would be a fairly straightforward hook to implement.

e.g, before:

```js
hook.data = {
  foo: '    not trimmed    ',
  bar: {
    baz: '   needs  trimming\t\n   '
  },
  bam: 'trimmed'
};
```

and after:

```js
hook.data = {
  foo: 'not trimmed',
  bar: {
    baz: 'needs  trimming'
  },
  bam: 'trimmed'
}
```

eddyystop:
This may be part of the forthcoming serialize hook.

eddyystop:
The `traverse` hook, now waiting for review for Auk, will handle this and much more.

https://github.com/feathersjs/feathers-hooks-common/pull/73

PR #73, opened 2016-12-12
Traverse and change hook.data .result .query or hook.anyObj

eddyystop:
- Issue #37 asks for hook to trim all strings
- BestBuy/api-playground has hook to convert all strings in .query that are "null" to null. Likely used with REST HTTP transports.
- This hook can do the above and much more.
- Hook uses substack/js-traverse which is reasonable in size.
- Can use this js-traverse's get/set methods in our get/setByDot utils instead of our own code.


ekryski:
Looks pretty good to me.

PR #76, opened 2016-12-18
Allows the include option to be a single object in populate hook

bertho-zero:
Allows the include option to be a single object

### DIFF
--- a/src/populate.js
+++ b/src/populate.js
@@ -41,8 +41,9 @@ export const populate = (options, ...rest) => {
           throw new errors.BadRequest('Schema does not resolve to an object. (populate)');
         }
 
-        return !schema1.include || !Object.keys(schema1.include).length ? items
-          : populateItemArray(options1, hook, items, schema1.include, 0);
+        const include = [].concat(schema1.include || []);
+
+        return !include.length ? items : populateItemArray(options1, hook, items, include, 0);
       })
       .then(items => {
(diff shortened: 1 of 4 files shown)


eddyystop:
Nice. Thanks.

Perhaps this should be implemented for `include` at any depth. The alternatives seem to be

- Add similar code at the end of `populateAddChild` so depths > 1 are handled.
- Handle all depths in `populateItemArray`, returning `items` if there is no `include`.

bertho-zero:
It's done, I added one test but phpstorm did anything with indentation..

eddyystop:
The files are being reorganized in feathers-hooks-common for event filter and permissions filters. I'm holding off on this PR until that settles down.

eddyystop:
First, sorry for the long wait on this PR. I was waiting on another PR that had to be pushed first to prevent merge conflicts arising.

The populate hook change looks great. Thanks!

The original tests tested `include: { {}, {}, ... }`. Those tests now appear to have been converted to only test for `include: {}` rather than to test for that in addition.

Since there's now a speed issue with the module reorganization, perhaps I should make a new PR containing your hook changes for the new module name. Plus I'll include your changed tests as new tests.

eddyystop:
See https://github.com/feathersjs/feathers-hooks-common/pull/88

bertho-zero , thanks for making the above trivial!

PR #83, opened 2016-12-23
Added missing dependency to package.json

eddyystop:


daffl:
I think we might end up with a circular dependency here since feathers-hooks pulls in this repository for providing the old bundled hooks.

daffl:
I think the issue is that it is being required in https://github.com/feathersjs/feathers-hooks-common/blob/master/src/new.js#L7. It's probably a good time to move `processHooks` into `feathers-commons`.

PR #84, opened 2016-12-23
Fixed 6.17.0 Babel transpiling error

eddyystop:
Using softDelete() pointed out a Babel 6.17.0 transpiling error in iffElse()


daffl:
What was the error? It looks like it should work (I just tried and it seems to be working with 6.20.0).

eddyystop:
`hooks => hooks ? combine(...hooks).call(this, hook)`

transpiled as

```javascript
var runHooks = function (hooks) {
  return hooks ? combine.apply(this, hooks).call(this, hook) : hook;
};
```

The second `this` is wrong.

daffl:
We can :shipit: I'd just be curious in which case it'd break (it looks right to me).

PR #88, opened 2017-01-06
Allowing populate include:{...} to act like include:[{...}]

eddyystop:
Don't require a single join to have to be specified as a 1 element array.

- Modified just the new module as the old will be removed shortly
- Relavent tests now run using both syntax

Recasting of PR #76 which has gotten dated and been closed.


eddyystop:
daffl says:
Looks good to me

ekryski:
Yup looks good!

PR #89, opened 2017-01-06
Working around another Babel-core 6.17.0 transpilation error in iffElse

eddyystop:
- Only new module changed as old will be deleted soon.
- They caused the  wrong 'this' being passed to predicate and hooks.
- @daffl says transpile is OK in Babel-core 6.21.0
- Added tests to catch any problems in predicate and hooks
- Babel-core 16.7.0 transpiles
const check = typeof predicate === 'function' ? predicate(...fnArgs) : !!predicate;
as
var check = typeof predicate === 'function' ? predicate.apply(undefined, fnArgs) : !!predicate;

### DIFF
--- a/src/common/conditionals.js
+++ b/src/common/conditionals.js
@@ -12,14 +12,15 @@ module.exports = function Conditionals (processFuncArray) {
     if (typeof trueFuncs === 'function') { trueFuncs = [trueFuncs]; }
     if (typeof falseFuncs === 'function') { falseFuncs = [falseFuncs]; }
 
-    // Babel 6.17.0 did not transpile something in the old version similar to this
-    // const runProcessFuncArray = funcs => processFuncArray.call(this, fnArgs, funcs);
+    // Babel-core 6.17.0 did not properly transpile the following when coded as an arrow func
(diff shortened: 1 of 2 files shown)


daffl:
What should the context be? In strict mode, the context of a function should be `undefined` when called like this. I'd also try and use the latest versions (clear cache and reinstall) in case it was a bug that has been fixed. I think Babel is around 6.20 now.

daffl:
I'm mainly asking because we might be fixing a non-issue. E.g. when using arrow functions at the module level scope, the context should always be `undefined` (which is why I don't like using top level arrow functions).

eddyystop:
The containing function is a top-level arrow functions. Thanks!

I have a bunch of PRs I'd like to push in order to avoid merge conflicts. So I suggest we push this PR for its tests and then I'll confirm the issue's with the top-level being an arrow function, as the details are a bit more complicated.

daffl:
Sounds good. 👍

ekryski:
Sounds good. :shipit: Let us know if that is still in an issue with Babel v6.20.

PR #90, opened 2017-01-06
Switching over to reorganized hook modules

eddyystop:
- Deleted original hook modules and tests
- Next step is to break new modules into 1 file per hook


daffl:
👍 Looking forward to the next step :) I think it'll help to make it more easy to navigate (and see what's happening).

ekryski:
:shipit:

PR #92, opened 2017-01-06
Break out filter and permission modules by hook

eddyystop:
PR 1
- The permission hooks were split.

PR 2

- Split filter hooks.
- Changed camelCase file names as per @daffl
- The PRs in pipeline along with this PR need to be pushed before I start service hooks, else there will be merge conflicts.
- common/conditionals-split & alter-data-split are temps till service done
- filters & permissions/conditionals.js will get some more attention


ekryski:
:shipit:

PR #94, opened 2017-01-07
Finish split of filters/ and permissions/. Prepare hooks/ for split.

eddyystop:
The commit comments indicate what's been done.


daffl:
I might've missed this in the previous PR but couldn't we put all hooks at the `src/` top level instead of separating them into folders? Lodash is doing the same thing and although it might be a lot of files it seems easier than having to find out what subfolder might be what you are looking for. Not a huge deal since most editors quick find should pull it up anyway but easier to e.g. see on GitHub all the hooks available.

eddyystop:
There have to be folders below src/ as hooks for filters, permissions and services have the same name but different logic.

ekryski:
Yup this looks good now! :shipit:

daffl:
Ah got it. 👍 then

PR #97, opened 2017-01-07
Split services hooks into individual files

eddyystop:
- All services, filters, permission hooks have been split into seperate files.
- Folders in `test` remain to be flattened.
- Docs have to be updated for client and validateSchema.
- Docs have to be updated as all services hooks now accessible via `require('feathers-hooks-common')`.

This needs to be pushed before I can work on some bugs (incl Babel) and issues.


ekryski:
:shipit:

PR #103, opened 2017-01-17
Populate 3

eddyystop:
Adding some enhancements to the populate hook. 2 of the 3 changes address issues people have raised.

One issue raised was that if the populate hook is on a service being called with a hook.params.provider, the calls for joining child services did not have a provider.

I checked other code, including the legacy populate hook with https://github.com/feathersjs/feathers-hooks-common/blob/master/src/services/legacy-populate.js#L36

They pass the parent's hook.param when calling the child. That certainly passes hook.params.provider to the child. I'm not yet familar enough to be certain that everything else that's passed along can't cause a problem. For example, what if the base service call included something with `.Model`?

So I'd like @daffl to review the first commit.


eddyystop:
daffl said ok in slack

Issue #28, opened 2016-10-31
disable hook seems to have wrong true/false logic


beeplin:
[Code here](https://github.com/feathersjs/feathers-hooks-common/blob/master/src/bundled.js#L336) means to disable when the result is false, not to disable when the result is true.
```js
if (typeof realm === 'function') {
    return function (hook) {
      const result = realm(hook);
      const update = check => {
        if (!check) {
          throw new errors.MethodNotAllowed(`Calling '${hook.method}' not allowed. (disable)`);
        }
      };
  }
```
But in [docs here](https://docs.feathersjs.com/hooks/bundled.html) it seems to be opposite logic: 

```js
  // Disable the remove hook if the user is not an admin
  remove: hooks.disable(function(hook) {
    return !hook.params.user.isAdmin
  })
```

Either the doc or the code is wrong. I guest it's the code which is wrong.

eddyystop:
Thanks for your example. I assume it should be `if (check) {` but let me ask what the ramifications might be regarding changing the code.

As an aside, a number of hooks, like disable, allow their last param to be a function. This function indicates if the hook should be run or not. That feature will be deprecated in all these hooks and removed in the future.

I recommend you use `hooksCommon.iff(condition, hook)` where condition may be sync, or async using a Promise. `iff(condition, hook1, hook2, ...)` will be supported in the future.

ekryski:
@eddyystop if it is wrong it's a bug and likely should be a patch release to feathers-hooks. However since it might cause a lot of breakage for people we **may** want to do a feathers-hooks major release. What do you think @daffl?

Regarding a release to `feathers-hooks-common` I think we can be safe to do a patch release as I don't know how many people have moved over, **or** we just follow the same process as with `feathers-hooks`.

eddyystop:
feathers-hooks now exports feathers-hooks-common as its bundled hooks. So
there is no moving over as such.

daffl:
We should probably just fix the docs then.

ekryski:
If that bug has been in there a while and it is now used by feathers-hooks there are likely people that might be depending on that functionality. I'm surprised it wasn't caught earlier...

The question is, should we fix this and issue a major release? I think so. Even though it is a bug and typically a patch release, this really is the wrong logic and would be a breaking change for people.

So I think the course of action is:

- As @daffl suggested we should update the docs to at least use the new recommended syntax.
- Fix and issue a major release for feathers-hooks-common
[...]

PR #108, opened 2017-01-18
Added disallow hook to start to deprecate disable hook.

eddyystop:
https://github.com/feathersjs/feathers-hooks-common/issues/28 we agreed to create a new hook rather than change disable.

- disable hook has a predicate & we are deprecating predicates.
- the disable predicate is implemented in perhaps the wrong sense.
- In https://github.com/feathersjs/feathers-hooks-common/issues/28 we agreed to create a new hook rather than change disable.


src/services/disable-method.js
```diff
@@ -0,0 +1,23 @@
...
+const errors = feathersErrors.errors;
+
+export default function (...providers) {
+  return hook => {
+    const hookProvider = (hook.params || {}).provider;
+
+    const anyProvider = providers.length === 0;
+    const thisProvider = providers.some(provider =>
+      provider === hookProvider ||
+      (provider === 'server' && !hookProvider) ||
```
daffl: The correct provider settings are:

- external: `!!provider`
- REST: `provider === 'rest'`
- Socket.io: `provider === 'socketio'`
- Primus: `provider === 'primus'`

README.md
```diff
@@ -24,8 +24,7 @@ will be removed next version. Use the new `populate` hook.
 - Use `deleteByDot` rather than `setByDot(obj, path, value, true)`.
 - The `delete` hook should be used instead of `remove`.
 You will need to wrap `delete` it in an conditional if you want it to work like `remove` does.
-- The `stop` hook may be used instead of `disable`.
-The `disable` hook when `false` instead of the documented `true`.
+- The `disableMethod` hook may be used instead of `disable`.
```
daffl: Should we maybe call it `disallow` or `refuse`?

eddyystop:
Both are much better names. I'd go with disallow to stay within the "dis"
family.

eddyystop:
I prefer names that are clear about what they do, even if they get long. I would prefer a name like `disallowProvider`. What do you think?

daffl:
I always think short names combined with the right parameters work. Something like `disallow('external')` or `disallow('rest')` is fairly explanatory to me. The `provider` postfix only makes sense if you know what a provider is (and I think @ekryski and @marshallswain wanted to change to calling it _Transport_).

ekryski:
Since we are already having to do a major version and have breaking changes, we may as well just call it disable and perform the functionality we want to. Rip that bandaid off instead of having even more legacy code to maintain. That's my 2 cents anyway.

ekryski:
Otherwise I agree with @daffl. Just go with `disallow`.

eddyystop:
Waiting for a ship it then.

ekryski:
:shipit:

daffl:
:shipit:

PR #110, opened 2017-01-19
Added disableMultiItemChange hook, throws if id===null for update, patch, remove

eddyystop:


daffl:
👍  Bit of a long name but the only shorter thing I could think of would be `disableMultiple`. No feeling super strongly about it so :shipit:

eddyystop:
This PR is basically a copy of something you posted, where you called it `disableMultiRecordChanges`. :)

PR #111, opened 2017-01-19
Added getWithSlug, modifyWithSlug hooks

eddyystop:
- get: getWithSlug('name'), patch: modifyWithSlug('name'), remove: modifyW...
- service.get(x) returns id=x || name=x via service.find({query:{$or: [...]})
- likewise patch(x, data), remove(x) via id=null && query:{$or: [...]}
- getWithSlug is not compatible with softDelete. modifyWithSlug is compatible.


src/services/get-with-slug.js
```diff
@@ -0,0 +1,30 @@
...
+export default function (slugName) {
+  return function (hook) {
+    checkContext(hook, 'before', ['get'], 'getWithSlug');
+
+    const idField = hook.service.id;
+    const query = {
+      $or: [{
+        [idField]: hook.id
+      }, {
+        [slugName]: hook.id
```
ekryski: You might want to also add `$limit: 1`.

ekryski:
Maybe rename to just `getByField`, `patchByField`, and `removeByField` just to keep them a little clearer? Ideally I'd love for this to just be 1 hook but I can't think of a good name without it becoming confusing.

eddyystop:
I like the name change. I can't get around to doing it before Fri 2pm'ish EST.

We can publish v3 if its needed right away and merge this over the weekend.

eddyystop:
@ekryski I like the idea of a single hook for this, working similarly to softDelete. I'd also consider allowing more than 1 non-id field. (Why not?) I'd also investigate an extra option so this hook would integrate with softDelete.

Ideas for a name: use/set/define + alternate/secondary + indices/fields

Only 1 i/o operation would be required per call, except for update would become a find plus update with `id`.

test/services/modify-with-slug.test.js
```diff
@@ -0,0 +1,103 @@
...
+        assert.equal(result.data.length, 5);
+      });
+  });
+
+  it('remove error using missing key', () => {
+    return user.remove('xxxx')
+      .then(() => {
+        assert.fail(true, false);
+      })
+      .catch(() => {});
```
daffl: I always `assert` that it's the expected error message for tests like that.

ekryski: True

daffl:
Why is `get-with-slug` not compatible with softDelete? Shouldn't the `find` hook add the `softDelete` criteria to the query either way?

eddyystop:
softDelete's get checks if the deleted flag is set. This does not.

I think softDelete should be used before this hook. softDelete should set
hook.params._softDelete with the field name being used for the delete flag.
Then this hook should see if that prop exists and change its query param.

I'm not feeling so great atm. (Nasty cigar after someone became a father
here.)

daffl:
But it also limits the `find` call to undeleted items (see https://github.com/feathersjs/feathers-hooks-common/blob/master/src/services/soft-delete.js#L22). So if the item is marked as deleted it will never be returned by a `find` and the slug method will throw a `NotFound` error.

eddyystop:
These hooks are copies of what we've presented several times as replies to people.

However I believe a different approach would be better. The signatures used here would be a special case, and integration with softDelete should be easier.

To persue that approach I need an answer to this: if the client codes `service.get({ a: 'a' })` does the server before hook get `hook.id==={ a: 'a' }`?

boxfish:
In mongoose adaptor, the `$or` query will throws 

> CastError: Cast to ObjectId failed for value "SLUG_NAME" at path "_id" for model "SERVICE_NAME"

 because the slug name is not a valid ObjectId string. This may be something that needs to be handled by the mongoose adapter.

eddyystop:
@daffl I don't want to reopen this issue in its entirity, but what do you think about boxfish's comment just above?

daffl:
I'm not sure I understand the problem. This query is being used for a `find`, why would it try to cast to an ObjectId? Also, I think this PR can be merged right?

daffl:
Ah that totally makes sense. @eddyystop Should we go back to dropping the `$or` and do a `find` with the slug, then if it doesn't find anything we just won't set `hook.result` which will run the original query? This will make two queries which we tried to avoid but it would work for everything.

eddyystop:
We have not merged this PR and these hooks don't exist. They are incompatible with softDelete and we were not comfortable with that limitation. I had an alternative approach in mind, but I'll have to remember what it was :( .

We can revisit this issue after we get Auk out and after we handle the common hook issues presently waiting on Auk being released.

PR #114, opened 2017-01-20
Bump dependencies; Use shx

eddyystop:
- Greenkeeper needed some dep versions bumped.
- Switch from rimraf to shx


daffl:
Looks good for the dependencies. I think the script is missing the `shx` command thought right (looking at [this](https://github.com/feathersjs/generator-feathers-plugin/blob/master/app/templates/package.json#L35))?

eddyystop:
Thanks for the shx catch! It would have taken a Windows dev to run into it.

PR #118, opened 2017-01-21
Dep remove

eddyystop:
- Copied in /src/utils.js#L17-L27
- Filter processing remains a copy of /src/events.js#L44-69
- Removed from package.json


eddyystop:
I'll change the copied promisify function to the one used by feathers-hooks-common at the same time we switch the filter processing to what feathers-socket-common surfaces.

ekryski:
:shipit:

PR #120, opened 2017-01-21
Added discard hook to deprecate the remove hook.

eddyystop:
- remove does not remove on server calls. discard removes always.
- Added deprecate msg to remove hook.
- Added note to migration section in README.
- remove is the last hook with some "special" feature inside.


ekryski:
:shipit:

PR #136, opened 2017-02-28
setting provider to undefined

### LINKED ISSUE #134: populate should set provider to null
### Steps to reproduce

1. Setup a hook on a service (ie. messages) that should only be run when `isProvider('external')`.
2. Populate entities from the service (ie. a channel's messages)

I believe it is [this line](https://github.com/feathersjs/feathers-hooks-common/blob/master/src/services/populate.js#L184) that needs to be changed.

### Expected behavior
External only hooks should not run

### Actual behavior
They do run

### System configuration
feathers-hooks-common@3.0.0-pre.1

ekryski:
### Summary

Pretty self explanatory. Fixes #134. Pretty hard to test. Would need to rework how the tests are setup pretty considerably I think.

### DIFF
--- a/src/services/populate.js
+++ b/src/services/populate.js
@@ -181,7 +181,7 @@ function populateAddChild (options, hook, parentItem, childSchema, depth) {
       if (paginateOption === true) { paginate = null; }
       if (typeof paginateOption === 'number') { paginate = { paginate: { default: paginateOption } }; }
 
-      const params = Object.assign({}, hook.params, paginate, { query, _populate: 'skip' });
+      const params = Object.assign({}, hook.params, paginate, { query, _populate: 'skip', provider: undefined });
       return serviceHandle.find(params);
     })
     .then(result => {


daffl:
How does this work with hooks that remove fields for external providers? E.g. populate a post with a user whose password is normally removed. In the populate user you will see the password.

ekryski:
@daffl true. I would argue you should probably be always removing the password hash (not just for external). Have a suggestion? It's broken if we don't do this, but might break other things.

daffl:
Well we can't always remove fields since e.g. the password is needed to be compared for authentication (which was why we were removing it only for external calls).

I don't have an easy fix for this right now but I was thinking that once we make hooks part of core and event emitting a hook we can add a `hook.dispatch` or similar property which always contains the data converted to a format that external users are allowed to see. This could then be used for dispatching events (we have the same problem there), populating related entities etc.

eddyystop:
I posted a comment in https://github.com/feathersjs/feathers-hooks-common/issues/134. The legacy populate did what the new one does. It know of at least 1 person's code what depends on it.

This PR would be a breaking change. It would

ekryski:
Ya, there are definitely implications. I still don't think the current implementation is doing the right thing, however to unblock me, @daffl suggested adding a before hook:

```js
(hook) => {
        if (hook.params._populate) {
          delete hook.params.provider;
        }
}
```

Which works, so we can stew on this a bit.

ekryski:
I think more discussion is needed on #134 but this PR is not the right solution. We can't take such a brute force approach here. Going to close.

PR #137, opened 2017-03-06
Avoided side-effect of client

eddyystop:
**Note: Don't comment before looking at 'semistandard fix' which is more than what the desc says.** I hadn't had a coffee yet.

- The pgm may use a "const" { query: { $client: {...}}} as the hook.params in several method calls.
- The client hook will modify this "const" object and it will not function as expected in subsequent method calls.
- This subtle effect may be time consuming to find.

### DIFF
--- a/src/services/client.js
+++ b/src/services/client.js
@@ -1,7 +1,7 @@
 
 export default function (...whitelist) {
   whitelist = typeof whitelist === 'string' ? [whitelist] : whitelist;
-  
+
   return hook => {
     const params = hook.params;
 
@@ -14,7 +14,7 @@ export default function (...whitelist) {
         }
       });
 
-      params.query = JSON.parse(JSON.stringify(params.query));
+      params.query = Object.assign({}, params.query);
       delete params.query.$client;
     }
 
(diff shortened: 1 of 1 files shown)


src/services/client.js
```diff
@@ -1,7 +1,8 @@
 
 export default function (...whitelist) {
+  whitelist = typeof whitelist === 'string' ? [whitelist] : whitelist;
```
daffl: This isn't an array every time?

eddyystop:
```js
const avoidHooks = { query: { $client: { _offline: { ignore: true } } } };
service.create(data1, avoidHooks); // works as expected
service.create(data2, avoidHooks); // doesn't work as avoidHooks was modified above
```

Yup, `whitelist` is always an array. Thanks.

eddyystop:
A potential gotcha remains once this change is merged. I'm not saying its an error, just that it may trip people up.

```js
const aQuery = { $client: { _offline: { ignore: true } } };
service.create(data1, { query: aQuery });
// aQuery = { $client: { _offline: { ignore: true } } } as expected
```

```js
const aParams = { query: { $client: { _offline: { ignore: true } } } };
service.create(data1, aParams);
// aParams = { query: {}, _offline: { ignore: true } } which may not be expected
```

daffl:
:shipit:

PR #146, opened 2017-04-04
Update populate hook to call toJSON if implemented

### LINKED ISSUE #144: Populate should throw upon detecting an ORM result.
A substantial percentage of populate issues come down to the DB not being confirmed to return plain JS objects. Mongoose and Serialize return ORM's particular to their DB.

The problem is not solved by having `populate` automatically convert these ORM's to JS objects because all hooks except JS objects. We could change `get-items.js` to convert the ORM's but Feathers in general expects JS objects.

We have decided its more generic to change the defaults for Mongoose and Sequelize to return JS objects, so a user has to explicitly op-out to return an ORM. I'm not sure if this change has been pushed yet but its not prevelant in the field.

To save developers' and core-members' time, `populate` should check for ORM results and throw an explanatory message.

DesignByOnyx:
Fixes #144 

I wholeheartedly believe this is the best solution. Please read my comments on the linked issue. I tested this with arrays of data (`GET /foo`) as well as with single items (`GET /foo/1`).

### DIFF
--- a/src/services/populate.js
+++ b/src/services/populate.js
@@ -110,6 +110,11 @@ function populateItem (options, hook, item, includeSchema, depth) {
         item._elapsed = elapsed;
       }
 
+      // #144 - Fixes issue with custom `toJSON` implementations (most ORMs)
+      if (typeof item.toJSON === 'function') {
+        item = item.toJSON();
+      }
+
       children.forEach(({ nameAs, items }) => {
         setByDot(item, nameAs, items);
       });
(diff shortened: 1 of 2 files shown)


eddyystop:
Thanks for taking the time to make this PR. Its obvious this issue is of importance to you.

However this does not solve the problem of ORMs either for hooks in general or for Feathers in general. It does not solve a similar problem for MongoDB.

Perhaps we can continue this discussion in https://github.com/feathersjs/feathers-hooks-common/issues/144#issuecomment-291428967 .

DesignByOnyx:
Closing this as the issue is bigger than what is solved here. For now, we can instruct ORM users to create an additional hook which calls `toJSON()` on the data.

PR #159, opened 2017-04-22
Add async schema validation support

TheBeastOfCaerbannog:
### Summary

This PR aims to add support of asynchronous schema validation. Example of async schema:
```
var Ajv = require('ajv');
var ajv = Ajv({allErrors: true});

var request = require("request");

ajv.addFormat('twitter', {
    async: true,
    validate: (userName) => new Promise((resolve, reject) => {
        request.get('https://twitter.com/' + userName, (err, res) => {
            if (err) reject(err);
            else resolve(res.statusCode == 200)
        });
    })
});

var schema = {
    "$async": true,
    "properties": {
        "twitter_username": {
            "type": "string",
            "format": "twitter"
        }
    }
};
```
Example from https://runkit.com/esp/ajv-asynchronous-validation

### Other Information

I guess _potentially_ it could break something, because tests were written in synchronous try-catch way. But Feathers.js executes hooks in async manner, so this should be fine 🤔

### DIFF
--- a/src/services/validate-schema.js
+++ b/src/services/validate-schema.js
@@ -30,11 +30,13 @@ export default function (schema, ajvOrAjv, options = { allErrors: true }) {
         // Handler for synchronous validation
         .then((isValid) => {
           if (!isValid) {
-            // eslint-disable-next-line prefer-promise-reject-errors
-            return Promise.reject({errors: validate.errors});
+            return Promise.reject(
+              new ajv.constructor.ValidationError(validate.errors));
           }
         })
         .catch((err) => {
+          if (!(err instanceof ajv.constructor.ValidationError)) throw err;
+
           invalid = true;
 
           err.errors.forEach(ajvError => {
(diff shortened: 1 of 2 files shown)


eddyystop:
The test module changes in the PR look like they check 'test succeeds unexpectedly' only if the validation throws instead of if it succeeds.

Further, texts involving the success and fail of a genuine async schema are needed.

I don't think I should review such a sensitive code change without first knowing that it looks like it works.  I can look at this further once the tests are improved.

TheBeastOfCaerbannog:
Oh, ok, sorry. Changed the schemas validation to less general, but backwards compatible way. Schema is considered async if it contains "$async: true" property, and in this case validateSchema returns promise. Otherwise schema validated in same way as before. 
Also covered async validation with tests.

eddyystop:
I think that's a good way to go. I review as soon as I can.

eddyystop:
@beeplin Would you mind commenting on this issue and code? Thanks.

beeplin:
didn't check details about the errors handling, but according to the new test, it should be working fine. And this new feature is definitely a must-have. :)

eddyystop:
This will be published with v3.2.1.

PR #165, opened 2017-05-02
Fixes 'homepage' in package.json

cpsubrian:
Since `npm docs feathers-hooks-common` opens the 'homepage', it ought to be correct :smile:


eddyystop:
Thanks!

PR #188, opened 2017-05-15
Pass all params to soft-delete's internal 'get' call

NikitaVlaznev:
Passing just provider is not enough, because it will trigger authorization in the internal call, and it will fail without params.headers. It is better to pass all original params, that will include the 'authenticated' flag. And it is also more correct to pass the original query to treat the result of the internal call as a final result.

### DIFF
--- a/src/services/soft-delete.js
+++ b/src/services/soft-delete.js
@@ -59,7 +59,15 @@ export default function (field) {
     }
 
     function throwIfItemDeleted (id) {
-      return service.get(id, { query: { $disableSoftDelete: true }, provider: hook.params.provider })
+      // It is important to pass all params, among which are:
+      // * 'provider' - is needed for succeeding hooks like 'restrictToRoles'.
+      // * 'authenticated' - flag, set by preceding 'authenticate' hook,
+      //    is required to skip 'authenticate', that will be called otherwise,
+      //    because the 'provider' is passed.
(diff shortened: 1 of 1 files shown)


eddyystop:
Good point.

I've never included a `query` in a `get`. I think you have to ensure `hook.params.query` exists before doing `hook.params.query.$disableSoftDelete = true;`.

I'm leary about including all of `hook.params`. That could, for example, include Mongoose or Sequelize specific parameters appropriate for the base record but not necessarily so for child records. My feeling is that its safer to explicitly include certain params props.

eddyystop:
I feel more comfortable making the changes included in https://github.com/feathersjs/feathers-hooks-common/pull/197 .

Does this handle you use case?
Can you see any typical cases it would not handle?

Thanks ... awaiting your reply.

eddyystop:
@NikitaVlaznev please feel free to reopen this PR if you have concerns with the PR mentioned above.

NikitaVlaznev:
> I feel more comfortable making the changes included in #197 .
> 
> Does this handle you use case?
> Can you see any typical cases it would not handle?
> 
> Thanks ... awaiting your reply.

Yes it handles perfectly. Great thanks!

eddyystop:
Related https://github.com/feathersjs/feathers-hooks-common/issues/217

Issue #186, opened 2017-05-13
Allow parents to enable populate on their direct children 


shawnlauzon:
### Steps to reproduce

I have a hook that depends on a resource to contain another resource, which is normally included via its own populate. Unfortunately, populate called by another populate is always skipped. Although this makes sense usually to prevent deeply embedded populates, without it I need to duplicate the populate schema from child resource into the parent resource.

Suggested form:
```javascript
const schema = { include: [
  service: ...
  parentField: ...
  childField: ...
  populateChild: true
]};
```

The `populateChild` field defaults to `false`. 

The current implementation uses the `_populate` field, but it is only checked for the value `skip`. This can continue, but only sets that field if `populateChild` is false (default).

Thus this code does not change: 
```javascript
if (hook.params._populate === 'skip') {
  // this service call made from another populate
  return hook;
}
```

Alternatively we can remove `_populate` altogether, but would like to hear from the community.

I'll create a PR if this seems reasonable.

### Expected behavior
If specified, the direct children of this hook have their children populated. This setting does not continue further down the stack, unless children specify it.

### Actual behavior
If a parent populates a relationship which naturally contain children, the parent must populate the grandchildren in the current hook.

eddyystop:
Implemented in https://github.com/feathersjs/feathers-hooks-common/issues/186 and will be included with v3.4.0.

Documented in https://github.com/feathersjs/feathers-docs/pull/617

PR #192, opened 2017-05-20
Added useInnerPopulate option to populate

eddyystop:
populate, when including records from a child service, ignores any populate hooks defined for that child service. The userInnerPopulate option will run those populate hooks.

This allows the populate for a base record to include child records containing their own **immediate** child records, with the populate for the base record knowing what those grandchildren populates are.

Resolves https://github.com/feathersjs/feathers-hooks-common/issues/186


eddyystop:
Resolves https://github.com/feathersjs/feathers-hooks-common/issues/186 and will be included with v3.4.0.

Issue #187, opened 2017-05-13
Make childField and parentField optional to support custom queries


shawnlauzon:
I would like to use populate to query fields which are related to the data coming in via a query rather than a parent / child relationship. For example, the resource coming in might be:

```javascript
{
  "portfolios": {
    "approvalStatus": "review"
  }
}
```
and I want my schema to look like:

```javascript
  include: [ {
    service: 'portfolios',
    select: (hook, parent) => parent.portfolios,
  }, ...
  // more here
```
In such a case there is no `parentField` or `childField`

Clearly if I only needed to return the portfolios, I would do `find?approvalStatus=review`. However, my service performs actions on the result of the select (in this case the portfolios are used to create another resource).

### Expected behavior
Treat `parentField` and `childField` as optional parameters

### Actual behavior
An error "Parentfield x for x depth 0 is undefined".

shawnlauzon:
Alternatively, I see now that if my `childField` is the same as the result of the `select` function, it will override it. However I don't have any fields that I always know will be there. I can fake one for the moment, but I'm wondering if some explicit behavior might be better.

eddyystop:
The `populate` hook does SQL-like joins, and the pattern for that is dependant on having a `key` and a `foreign key` to relate the related records.

(1) I'm still not clear what use case you are describing. Could you provide an example with sample source table records, the query parms coming in, and the desired result records.

(2) Its possible to dynamically generate schemas for `populate` as the `schema` property may be a function. Can you generate an SQL join relationship from your query params?

shawnlauzon:
Does `populate` need to be restricted to SQL-like join semantics? If so, then requiring a key and foreign key is the correct approach. 

However, I would like to use `populate` as a declarative query mechanism rather than specifically for SQL-like joins. Most of the functionality is the same and useful; it's the addition of the join functionality which necessitates the key and foreign key.

My use case is to create a new resource using query parameters coming from the client to define its contents, and then allow the feathers db adapter to store it. I would like to:

1. `populate` based on query parameters coming in via the POST
2. Transform this data using a custom hook to the format I want to save
3. Database stores new resource
4. New resource returned to user.

My current workaround looks like this:
```javascript
  include: {
    service: 'api/v0/portfolios',
    nameAs: 'portfolios',
    select: (hook, parent) => Object.assign({}, { shortId: { $exists: true } }, parent.portfolios)
    parentField: 'portfolios',
    childField: 'shortId'
}
```

[...]

PR #194, opened 2017-05-20
Enhanced populate so parentField and childField are optional

eddyystop:
- Populate has joined child records to a parent record using the related columns
parentField and childField.
- However populate's query and select options can be used to related the
records without needing to use the related columns.
- This is a more flexible, non-SQL-like way of relating records.
- It easily supports dynamic, run-time schemas since the select option may be
a function, i.e. (hook, parentItem, depth) => queryObj.


eddyystop:
Resolves https://github.com/feathersjs/feathers-hooks-common/issues/187 and will be published as v3.4.0.

Issue #163, opened 2017-05-01
softDelete fix for double 'get' call is not ideal


NikitaVlaznev:
softDelete fix #160 have one more issue.

Currently it uses internal users.get result as final result. It's ok except that internal calls are made without hook.params.provider set, making some other hooks to be skipped, like restrictToRoles.

Thus, if restrictToRoles hook is registered after softDelete it will not be called at all, it is a security issue.

Somehow softDelete should call users.get like an external service, could someone give an advice about how to do it?

I see that it can be fixed by adding softDelete as after get hook, and by making softDelete before get hook just pass through with just setting some flag like $disableSoftDelete, but it will brake compatibility, everyone will need to add softDelete to after get hooks, may be there is a way that a hook can use to register after hook by it self?

Btw, current tests does not see that issue.

eddyystop:
I believe this situation was introduced by your https://github.com/feathersjs/feathers-hooks-common/pull/160. I see 2 solutions
- Change `return service.get(id, { query: { $disableSoftDelete: true } })` to `return service.get(id, { query: { $disableSoftDelete: true }, provider: hook.params.provider })
- Revert https://github.com/feathersjs/feathers-hooks-common/pull/160 and do 2 `get`s.

What are the ramifications of the first approach above? Some hooks could prevent that `get` from being successful.

softDelete is designed to be an all-before hook. I think its unwise to break it up into before & after hooks.

NikitaVlaznev:
This `service.get(id, { query: { $disableSoftDelete: true }, provider: hook.params.provider })` is exactly what i was looking for, if service methods really can accept provider then it is the solution i think.

NikitaVlaznev:
I think if some hooks can prevent that get, then it should be prevented.

What hooks can prevent it except access controlling hooks?

eddyystop:
Fixed in https://github.com/feathersjs/feathers-hooks-common/pull/166 which will be published in v3.2.1.

3bola:
Now when you call softDelete on users service before all hook, and on client side straight after authentication you try to get the user, this will fail.
This could be fixed by including hook params user and authenticated in the `service.get(id, { query: { $disableSoftDelete: true }, provider: hook.params.provider })` call.

You can try this error out at my example repo @ https://github.com/3bola/example-feathers-soft-delete-errors

eddyystop:
Nice work!
@ekryski ^^^

NikitaVlaznev:
[...]

Issue #185, opened 2017-05-11
There may be an interaction between softDelete and authentication


eddyystop:
Someone reported authentication working with softDelete in feathers-hooks-common v3.2.0 but not v3.3.0.  The softDelete was on the `users` service.

@ekryski is tracking the issue down, and is not positive the problem is with softDelete.

softDelete used to do 2 `get`s for `service.get`, one with `provider: undefined`, the other with `provider: hook.para,s.provider`. Now it does just one with `provider: hook.params.provider`.

eddyystop:
More info https://github.com/feathersjs/feathers-hooks-common/issues/163#issuecomment-300752881

johncrisostomo:
I don't know if this is related, but I was trying out softDelete and it works as expected, the only problem I'm getting from it is it's still giving me 
```
{
  "name": "NotFound",
  "message": "Item not found.",
  "code": 404,
  "className": "not-found",
  "errors": {}
}
```
Even though it was able to successfully patch the item with `deleted: true`. The before hooks from this service looks like this:
```
  before: {
    all:  [ softDelete() ],
    find: [],
    get: [],
    create: [ 
      authenticate('jwt'),
      associateCurrentUser({ as: 'ownerId' }),
    ],
    update: [...restrict],
    patch: [...restrict],
    remove: [...restrict],
  },
```

eddyystop:
Am I correct in understanding you got the NotFound message during the `patch` call, even though that call changed the record as expected?

`softDelete` does not contain any after hooks, so it ran nothing after the successful `patch` call.

That would be the message you'd get if you tried a `get` after patching the record to be deleted.

Perhaps you can clarify the situation, or provide some code that shows the situation?

johncrisostomo:
@eddyystop Hi eddy. Yeah sure, it's actually a new toy project, I'm familiarizing myself with Auk so I can update some of our legacy projects that were using older versions/styles. The code is mostly boilerplate code yet from the latest `feathers-cli`. And no, I'm not trying to do a `get` after the remove. I just send a delete request from Postman and that's what I get. 

https://github.com/johncrisostomo/voting-backend/tree/master/src/services/polls

eddyystop:
I couldn't find a `softDelete` in either `polls` or `users` in the link above. Could you provide a breaking example and instructions to cause it?

Could also run with `feathers-hooks-common` v3.2.0 rather than the current one? Thanks.

johncrisostomo:
[...]

PR #197, opened 2017-05-21
Resolved issues in softDelete caused by pull #163

eddyystop:
- https://github.com/feathersjs/feathers-hooks-common/issues/163 eliminated an unneeded `get` when doing a softDelete `get` call.
- However issues were introduced with authentication
- This is a focused fix which may be safer than the one proposed in pull https://github.com/feathersjs/feathers-hooks-common/pull/188


eddyystop:
Might also affect https://github.com/feathersjs/feathers-hooks-common/issues/185

PR #202, opened 2017-05-29
add missing `feathers-hooks` dependency

ahdinosaur:
`feathers-hooks` is used here: https://github.com/feathersjs/feathers-hooks-common/blob/8c1ab2fa228a2e2e42c311f48c99488900064296/src/services/combine.js

this change fixes an error when using `feathers-hooks-common` in a standalone module without `feathers-hooks` as a dependency.


eddyystop:
Thanks for the catch!

PR #212, opened 2017-06-16
fix: pin ajv to 5.1.5

eddyystop:


eddyystop:
I'll unpin to move to ajv 5.1.6 when they have fixed their issue.

PR #232, opened 2017-07-27
Added hook utility thenifyHook

eddyystop:
Running an after hook conditionally can increase the cognative load. Consider running a `populate` conditionally:
```javascript
// in one place
service.get(id,{ _includeTeams: true });
// and in a separate module
after: { get: iff(context => context.params._includeTeams, populate(...)}
```

`thenifyHook` allows after hooks to run in the `.then` chain of the service call. The above can now be written as:
```javascript
const thenify = thenifyHook({ app }); // props to add to the hook call's `context`
service.get(id).then(thenify(populate(...)));
```
The intent is clearer.

`thenifyHook` can be used with any after hook once you set up the hook's context correctly.

### DIFF
--- a/src/services/index.js
+++ b/src/services/index.js
@@ -37,6 +37,7 @@ import setUpdatedAt from './set-updated-at';
 import sifter from './sifter';
 import softDelete from './soft-delete';
 import stashBefore from './stash-before';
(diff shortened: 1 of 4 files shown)


eddyystop:
Published as 3.6.0.

PR #233, opened 2017-07-27
Made `populate` hook friendlier to `thenifyHook` util

eddyystop:
- if `schema.service` specifies the base service name, populate will not
throw if context.path isn't there.

So you don't have to include `path` when using `thenifyHook`.


eddyystop:
Will be released with >= 3.6.1

PR #234, opened 2017-07-27
Fixed 2 issues with validateSchema

eddyystop:
- `options.addNewError` was needlessly deleted from the `option` param, modifying that input param.
- The ambiguous `Invalid schema` error.BadRequest text was replaced with the clearer `Data does not match schema`


eddyystop:
published as 3.6.1

PR #239, opened 2017-08-03
Add top level provider option to populate hook

adamvr:
### Summary

This pull request extends the populate hook to allow for the specification of a global provider for the entire populate tree, rather than specifying them on a per child schema basis. The motivation here is that it's quite easy to miss a provider somewhere in your schema tree, particularly if it is large.

In cases of nested includes, the provider will be inherited from the parent schema. If not specified, the provider will default to the hook's provider.

Example: 

```javascript
const schema = {
  service: 'users',
  nameAs: 'user',
  parentField: 'userId',
  childField: 'id',
  // undefined / server provider will be used unless overriden by a child schema
  provider: undefined,
  include: {
    service: 'notifications',
    nameAs: 'notifications',
    parentField: '_id',
    childField: 'userId',
    // `notifications` service will be accessed with the `rest` provider
    provider: 'rest'
  }
};

populate({ schema })
```

### DIFF
--- a/src/services/populate.js
+++ b/src/services/populate.js
@@ -24,7 +24,8 @@ export default function (options, ...rest) {
     const optionsDefault = {
       schema: {},
       checkPermissions: () => true,
-      profile: false
+      profile: false,
+      provider: hook.params.provider
     };
 
     if (hook.params._populate === 'skip') { // this service call made from another populate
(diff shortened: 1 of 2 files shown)


eddyystop:
What do you think about moving that provider option to the root of the schema object? It seems a more appropriate place for it.

```javascript
const schema = {
  service: '...',
  permissions: '...',
  provider: '...',
  include: [ ... ]
};
```

adamvr:
Yep, makes sense. Updated.

eddyystop:
Thanks!

Published as v3.7.0.

adamvr:
Awesome!

PR #246, opened 2017-08-21
Fix _include being overwritten with empty array

rodeyseijkens:
Fix _include being overwritten with empty array so multiple / external populate functions can add to the _include too

**Problem before fix:**
I was writing a custom populate hook function and I wanted to have it also add it to the _include array so all the code is neat and consistent, but when I do another populate hook from feather after that it will set the array to empty so it will be removed, this will check if there is already an _include array before creating an empty one.

### DIFF
--- a/src/services/populate.js
+++ b/src/services/populate.js
@@ -103,7 +103,7 @@ function populateItem (options, hook, item, includeSchema, depth) {
   const elapsed = {};
   const startAtAllIncludes = process.hrtime();
   const include = [].concat(includeSchema || []);
-  item._include = [];
+  if(!Object.prototype.hasOwnProperty.call(item, '_include')) item._include = [];
 
   return Promise.all(
     include.map(childSchema => {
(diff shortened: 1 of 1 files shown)


eddyystop:
Nice. I did not consider multiple populate hooks.

The same issue will occur with `_computed`. Would you care to handle that?

I don't think anything can be done to have a consolidated `_profile`.

The PR fails on 2 tests. Please update them
```text
  services populate - 1:1 & 1:m & m:1 - array
    ✓ for one item
    1) for an item array
  services populate - 1:1 & 1:m & m:1 - obj
    ✓ for one item
    2) for an item array
```

rodeyseijkens:
Looks like everything is fixed for this now.

I looked into the _computed but as far as I can tell it already does that?¿

eddyystop:
Thanks!

eddyystop:
Published in v3.7.2

PR #251, opened 2017-09-06
Allow schema IDs in first parameter of validateSchema

CypherAlmasy:
### Summary

- [X] Tell us about the problem your pull request is solving.

This is not so much a problem as it is a convenience.  Users can optionally provide a schema ID for a pre-loaded schema in an ajv instance, rather than having to pull in the schema itself from elsewhere to set up the hook.  This helps ensure readable code and promotes schema re-use.

- [ ] Are there any open issues that are related to this?   -    No
- [ ] Is this PR dependent on PRs in other repos?    -   No

### Other Information

No other information I can think of which is relevant; it's a pretty small change.  I'm, of course, totally willing to submit a relevant PR to feathers-docs if this is merged.

### DIFF
--- a/src/services/validate-schema.js
+++ b/src/services/validate-schema.js
@@ -16,7 +16,7 @@ export default function (schema, ajvOrAjv, options = { allErrors: true }) {
   } else {
     ajv = ajvOrAjv;
   }
-  const validate = ajv.compile(schema); // for fastest execution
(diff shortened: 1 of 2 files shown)


eddyystop:
Nice addition. I'll be released the next time this repo is published to npm.

PR #253, opened 2017-09-12
no need for populate in softDelete

superbarne:
### Summary
softDelete checks only if (data[deleteField]) so there is no need this query should be populated.


eddyystop:
This is a nice catch. Thanks!

This'll be with the next version published.

eddyystop:
Published in v3.7.3

PR #259, opened 2017-09-29
soft-delete 'setDeleted' feature

BenZed:
soft delete can now take a string or an object as configuration,
exposing an additional field: setDeleted

setDeleted allows you to specify a function that returns a value that
will be used when a record is soft deleted

use case examples would be a date object or the id of a user that
triggered the deletion.

setDeleted checks to ensure the value returned is truthy, and may be
asynchronous.

Other notes:
- tests refactored, no longer throw promise warnings.
- removed deprecated babel-preset-es2015 in favour of babel-preset-env.
I'm aware that feathers.js is planning on removing babel, but for the
time being this prevents another deprecation warning.
- soft-delete now tries to look for a custom authentication entity
field before defaulting to 'user' in the 'throwIfDeleted' subroutine.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,7 @@
 {
-  "plugins": [ "add-module-exports" ],
-  "presets": [ "es2015" ]
+  "presets": [ "env" ],
+  "plugins": [
+    "add-module-exports",
+    "transform-runtime"
+  ]
 }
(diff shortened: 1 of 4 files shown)


BenZed:
(made a mistake, i'll re submit later)

PR #260, opened 2017-09-29
Soft delete feature

BenZed:
The `softDelete` hook can now take a `string` or an `object` as configuration.

Providing an object allows additional fields:

### `config.setDeleted`
`setDeleted` allows you to specify a function that returns a **value** that
will be used when a record is soft deleted.

Use cases:
  - A **Date** object, denoting when the Record was deleted.
  - A `user.id`, denoting who deleted the Record.

An `Error` will be thrown if the value returned from `setDeleted` is not *truthy*.

### `config.disableParam`
Customizes the name of the disable param from `$disableSoftDelete` to
something else.

### `config.allowClientDisable`
If false, the hook will ignore the `$disableSoftDelete` param on the
**query** if it's been provided by a **client**.

---

All of these new options are configured with defaults that would make the behavior of the `softDelete` hook consistent to previous releases.

#### Other notes:

- Tests refactored, no longer throw promise warnings.

- `$disableSoftDelete` behavior changed. In previous releases, the `$disableSoftDelete` parameter was deleted from the `query` object one it hit the `softDelete` hook.  It is now moved from the `query` object to the `params` object. This way, if future hooks need to tell if something has been permanently deleted, they can reference the `params` object rather than making *additional database calls*.

- `softDelete` now tries to look for a custom *authentication entity field* before defaulting to **user** in the `throwIfDeleted` function. It will default to *user* if it can't find one.

- Removed deprecated *babel-preset-es2015* in favor of *babel-preset-env*.
I'm aware that **feathers.js** is planning on removing babel, but for the
time being this prevents another deprecation warning.


eddyystop:
I'll look into using your ideas here the next time I work on this hook.

eddyystop:
`softDelete2` is a noteable improvement over `softDelete`. softDelete2 does not run the `get` and `patch` hooks on its probing get and removal patch calls. It play correctly with authentication and its restrictToOwner hook.

- config.disableParam: Changing semaphore names is problematic since common hooks do in fact check the semaphores set by other hooks. So it could happen another hook needs to check if its running because softDelete2 made a service call. We have also started to standardize these semaphore names and how they will be stored in context.params.

- config.allowClientDisable: softDelete2 has the `allowIgnoreDeletedAt` option.

- config.setDeleted: softDelete2 allows a custom patch function for logically removing records. You can affect the record however you want within it.

-$disableSoftDelete in query. softDelete2's uses params.$disableSoftDelete2.

- authentication entity field: softDelete2 allows a custom get function for checking if the record is active. You can throw depending on the contents of the record.

- babel: Transpiling of the common hooks was removed by error when Buzzard was released. This will be fixed.

I think this issue can now be closed.

PR #266, opened 2017-10-19
Keep hook

Mattchewone:
### Summary

Created a whitelist keep hook. Rather than specifying fields you wish to remove, here you can specify fields you wish to keep.

### DIFF
--- a/src/services/discard.js
+++ b/src/services/discard.js
@@ -5,7 +5,7 @@ import getItems from './get-items';
 
 export default function (...fieldNames) {
   return hook => {
-    checkContextIf(hook, 'before', ['create', 'update', 'patch'], 'delete');
+    checkContextIf(hook, 'before', ['create', 'update', 'patch'], 'discard');
 
     _remove(getItems(hook), fieldNames);
 
--- a/src/services/index.js
+++ b/src/services/index.js
@@ -16,6 +16,7 @@ import existsByDot from '../common/exists-by-dot';
 import getByDot from '../common/get-by-dot';
 import getItems from './get-items';
 import isProvider from './is-provider';
+import keep from './keep';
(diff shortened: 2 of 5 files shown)


Mattchewone:
So it looks like the issue why this is failing is because on the latest version of Node `feathers-memory@1.0.1` is being installed, but on the older versions of Node `feathers-memory@1.2.1` is being installed.

In the later versions of `feathers-memory` there was a change to how the `$select` works from a custom method to using `feathers-common` select method, within that method the `id` is being added to the custom `$select`.

I have raised an issue with `feathers-memory` here - https://github.com/feathersjs-ecosystem/feathers-memory/issues/58

eddyystop:
Thanks for creating that issue. daffl is the go-to person for that.

eddyystop:
This looks good. I think its the very first third party hook added. Thanks.

However you have to also update the Feathers docs in feathers-docs. I'll merge both PRs together.

Mattchewone:
Thanks @eddyystop.

I have just created a PR for the docs here - https://github.com/feathersjs/feathers-docs/pull/879

eddyystop:
Nicely done. Thanks.

eddyystop:
Published in 3.10.0

bertho-zero:
`keep` is just the `pluck` hook without magic logic, no ?

eddyystop:
Yes, like discard (no magic) vs remove (magic). The docs will be updated to highlight the feature/danger.

bertho-zero:
All hooks with magic logic should be deprecated, no ?

eddyystop:
We started that with the last version of Feathersjs. We're now moving into "sensitive" area where a misreading of the docs could cause data loss for server-initiated service calls.

PR #274, opened 2017-10-27
Update to new plugin infrastructure

daffl:
As proposed in https://github.com/feathersjs/feathers/issues/608

Drops support for Node 4 but makes it possible to install from Git directly without having to run a compilation step and removes all Babel dependencies.

I think the large diff is because the code transformation changed all the linefeeds. Only the import statements really changed.


eddyystop:
I'll get on this as soon as I have time.

eddyystop:
I'm still having sys admin issues, so I'm merging this blind. :)

PR #279, opened 2017-11-06
Remove deprecated hooks

daffl:
Follow-up on #278 since this should be a major version change we can also remove the hooks that already print a deprecation message.


eddyystop:
Thanks.

As an aside, I'll soon be adding a new populating hook which runs roughly an order of magnitude faster than `populate`.

I'm also rewritting the docs for all the hooks.

ekryski:
@eddyystop that sounds awesome! Where were you planning on putting the docs updates? In `feathers-docs`, this repo, or https://github.com/feathers-plus/docs.feathers-plus.org?

Asking because we're planning on announcing the v3 updates and repo changes in the next week or so and then will likely be prioritizing some docs updates.

eddyystop:
I think the Hexo look is a great improvement over GitBook.

The theme feathers-plus uses is a modification of Vuejs' theme, and porting GitBook pages would be straightforward. Let me know if you want to consider switching to Hexo

daffl:
Yes. I definitely think switching should be considered. It looks like GitBook totally dropped supporting their self-hosted open source platform which is fair but makes it not a long-term solution for us.

ekryski:
Pasting an image for visibility...

image

---

Ya I think that looks pretty good. The Gitbook styles were inspired by Vue already. I had looked at Hexo before but maybe the next website revision and docs could switch over to that. It's really not a whole lot different than what the main Feathers website uses already.

In any case, we'll keep an eye on things here but @eddyystop if you can let us know when you are close to publishing that would be great.

@daffl once this one goes live it might make sense to pull the common hooks from the main docs and just link to this book.

eddyystop:
I published the website to GitHub with one command, and its served at https://feathers-plus.github.io . I think that is a very elegant feature.

PR #281, opened 2017-11-14
Added fastJoin and related makeCallingParams plus tests

eddyystop:


eddyystop:
Tests using async/await are failing Node 6. I'm mergin anyway as people are waiting on fastJoin. I assume I'll have to condition the tests by Node version.

PR #283, opened 2017-11-14
Deprecated pluck and client hooks

eddyystop:


daffl:
The next version this is going to be published with has breaking changes anyway, why not just remove them?

eddyystop:
I was also planning to
- write keepQuery, deprecate pluckQuery
- write discardQuery, deprecate removeQuery
- deprevate setCreateAt and setUpdateAt for setNow('createdAt') and setNow('updatedAt')

They would all have to be manually changed.

eddyystop:
I've thought about this. Changing Feathers core is pretty much an all or nothing thing. It doesn't make much sense to deprecate features.

I feel people will be challenged enough (time-wise, etc.) converting to the new Feathers core. I also feel its a good idea to not require them to handle hooks that disappeared without warning before they can even start testing the new core on their app.

I dislike old stuff lying around but I'll do it to help the users.

PR #289, opened 2017-11-18
fix feathers versions

bertho-zero:


eddyystop:
You know, I'm not sure if Auk is v2 or v3. v2 may have came out in July 2016 after which Feathers started to take off.

So I'm not going to use versions, just the bird names.

PR #300, opened 2017-11-22
Added option to preventChanges to remove invalid prop names rather th…

eddyystop:


eddyystop:
In Buzzard version https://feathers-plus.github.io/v1/feathers-hooks-common/index.html#preventchanges

PR #302, opened 2017-11-23
Added disablePagination hook.

eddyystop:
if context.query.$limit === -1 or '-1' then disable pagination


eddyystop:
Added in Buzzard https://feathers-plus.github.io/v1/feathers-hooks-common/index.html#disablePagination

PR #303, opened 2017-11-23
Added required hook to check fields dot notation.

eddyystop:
Must exist, not be falsey, numeric 0 is acceptable.


eddyystop:
Added to Buzzard https://feathers-plus.github.io/v1/feathers-hooks-common/index.html#required

PR #304, opened 2017-11-23
Added runParallel hook.

eddyystop:


eddyystop:
Added in Buzzard https://feathers-plus.github.io/v1/feathers-hooks-common/index.html#runParallel

PR #316, opened 2017-12-13
Modify keep hook to only ignore undefined values.

otang:
### Summary

Problem: keep is discarding properties if they have "falsey" values.

Solution: keep now only discards undefined values. As a result a 0 or empty string can now reach the service.

### DIFF
--- a/lib/services/keep.js
+++ b/lib/services/keep.js
@@ -1,6 +1,7 @@
 const checkContextIf = require('./check-context-if');
 const getItems = require('./get-items');
 const replaceItems = require('./replace-items');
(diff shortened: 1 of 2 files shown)


eddyystop:
Thanks for this catch. Its pretty serious.

Do you think it would be better to use `existsByDot` to see if the property exists and copy it regardless of its value?

otang:
@eddyystop Yep good idea. I've updated the PR to use `existsByDot` and modified the tests to check falsey values.

eddyystop:
Very nice. Thanks a lot!

PR #325, opened 2017-12-29
Verify deprecated syntax outside the hook function

guillaumerxl:
Hi, 
I think it's better to set `fieldNames` outside the hook function. The current version makes `fieldNames[0]` to always be a string after `fieldNames = fieldNames.slice(1);`. So even if the developer uses the `preventChanges(true, ...fieldNames)` syntax, the hook will log the deprecated warning.
With this solution, the warning will be logged only if the developer is using the deprecated syntax.

### DIFF
--- a/lib/services/prevent-changes.js
+++ b/lib/services/prevent-changes.js
@@ -5,15 +5,19 @@ const checkContext = require('./check-context');
 const errors = require('@feathersjs/errors');
 
 module.exports = function (...fieldNames) {
+  const ifThrow = fieldNames[0];
+  const deprecated = typeof ifThrow === 'string'
+
+  if (!deprecated) {
+    fieldNames = fieldNames.slice(1);
+  }
+
   return context => {
     checkContext(context, 'before', ['patch'], 'preventChanges');
     const data = context.data;
-    const ifThrow = fieldNames[0];
 
-    if (typeof ifThrow === 'string') {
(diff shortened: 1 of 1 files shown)


eddyystop:
I think the difference is
1. The current code logs a message if the hook is **coded** with the deprecated signature.
2. Your change logs a message if the hook is **used** with the deprecated signature.

I think its better to find errors asap. I would rather see a message during development when the server is started up, than depend on finding it in a log later on, possibly in production.

Also, the code handing the deprecated signature will be removed in the next major version, likely causing that signature to cause unexpected results or to crash the hook. So its important to identify what code needs to be reworked.

What are your thoughts?

guillaumerxl:
Hello, 
it seems the current code logs a message even if the hook is not coded with the deprecated signature. This is what I'm experiencing. 
But I agree with you about identifying the problem during the startup of the server. 

So we just need to log the deprecated message outside of the hook function instead of logging it in each hook call?

eddyystop:
I will look into this when I return from vacation next Wednesday.

eddyystop:
I'll be adding a PR to log the deprecated message in the outer function.

Thanks for this.

eddyystop:
Repo updated https://github.com/feathers-plus/feathers-hooks-common/pull/329

It will be published to npm as 4.1..1 or 4.2.0.

PR #332, opened 2018-01-13
Improve alterItems hook: can map to a newly created object

beeplin:
Previously `alterItems` only accept a function that use `item` as a reference, and we have to create/mutate/delete properties. upon the original reference This makes a bunch of `lodash` tools unusable, since most `lodash` functions return a new object rather than changing the original input.

This PR allows `alterItems` accepting a function with a returned value. If the function returns a non-null value, the original item/items are replaced with the returned value; if the function returns nothing, just keep the previous behavior. 

Example: 

```js
alterItems(item => {
  return _.omit(item, ['a', 'b', 'c'])
}
```
Please check the code and the tests with caution in case I make some terrible mistakes. ;)

### DIFF
--- a/lib/services/alter-items.js
+++ b/lib/services/alter-items.js
@@ -18,7 +18,7 @@ module.exports = function (func) {
     (isArray ? items : [items]).forEach(
       (item, index) => {
         const result = func(item, context);
-        if (result != null) {
+        if (typeof result === 'object' && result !== null) {
           if (isArray) {
             items[index] = result;
           } else {
--- a/lib/services/alter-items.js
+++ b/lib/services/alter-items.js
@@ -13,8 +13,20 @@ module.exports = function (func) {
   }
 
   return context => {
-    const items = getItems(context);
-    (Array.isArray(items) ? items : [items]).forEach(item => { func(item, context); });
+    let items = getItems(context);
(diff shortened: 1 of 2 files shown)


beeplin:
I realized this is a breaking change, and maybe we should create a new hook for this purpose?

eddyystop:
Is it really a breaking change? The current version ignores any return value and the docs state "Function modifies item in place."

I would prefer using `if (typeof result === 'object' && object !== null)`.

I might have to delay updating the docs for 1 or 2 weeks as I'm focused on something else.

eddyystop:
`doSomethingForItem` would have to return something to have a  problem. I could agrue that's bad code; that people get burned when playing with edge cases.

`alterItems` has not be out very long so the 'install base' is small. Buzzard is not yet released. V4.x.x is not yet published to npm.

I would rather not not create YAH (yet another hook).

beeplin:
Agree. Me too hate YAH :))

So I will take your `if (typeof result === 'object' && object !== null)` and update the PR.

beeplin:
done.

eddyystop:
https://github.com/feathers-plus/docs/issues/10

I'll wait to bump the semver.

PR #333, opened 2018-01-13
Improve keep hook: allow passing in a function (item, context) => fields

beeplin:
Sometimes we need to dynamically determine which fields to keep according to the current item and context. This PR allows the `keep` hook to accept a function to dynamically generate field names. Example:

```js
keep((item, context) => 
  context.method === 'update' ? item.fieldsToKeep : ['_id', 'name']
)
```

### DIFF
--- a/lib/services/keep.js
+++ b/lib/services/keep.js
@@ -5,20 +5,28 @@ const existsByDot = require('../common/exists-by-dot');
 const getByDot = require('../common/get-by-dot');
 const setByDot = require('../common/set-by-dot');
 
-module.exports = function (...fieldNames) {
+module.exports = function (...fieldNamesOrFunc) {
   return context => {
     checkContextIf(context, 'before', ['create', 'update', 'patch'], 'keep');
     const items = getItems(context);
 
     if (Array.isArray(items)) {
-      replaceItems(context, items.map(item => replaceItem(item, fieldNames)));
+      replaceItems(context, items.map(item => replaceItem(item, getFieldNames(fieldNamesOrFunc, item, context))));
     } else {
-      replaceItems(context, replaceItem(items, fieldNames));
+      replaceItems(context, replaceItem(items, getFieldNames(fieldNamesOrFunc, items, context)));
     }
 
     return context;
   };
 };
+
+function getFieldNames (fieldNamesOrFunc, item, context) {
+  if (typeof fieldNamesOrFunc[0] === 'function') {
+    const result = fieldNamesOrFunc[0](item, context);
+    return Array.isArray(result) ? result : [result];
+  }
+  return fieldNamesOrFunc;
+}
 
 function replaceItem (item, fields) {
   const newItem = {};
--- a/tests/services/keep.test.js
+++ b/tests/services/keep.test.js
@@ -136,6 +136,98 @@ describe('services keep', () => {
       hooks.keep('first')(hook);
       assert.deepEqual(hook.data, { first: '' });
     });
+
+    it('updates hook before::create, with function', () => {
+      hooks.keep((item, context) => 'last')(hookBefore);
+      assert.deepEqual(hookBefore.data, { last: 'Doe' });
+    });
+
+    it('updates hook after::find with pagination, with function', () => {
+      hooks.keep((item, context) => 'first')(hookFindPaginate);
+      assert.deepEqual(hookFindPaginate.result.data, [
+        { first: 'John' },
+        { first: 'Jane' }
+      ]);
+    });
+
+    it('updates hook after::find with no pagination, with function', () => {
+      hooks.keep((item, context) => 'first')(hookFind);
+      assert.deepEqual(hookFind.result, [
+        { first: 'John' },
+        { first: 'Jane' }
+      ]);
+    });
+
+    it('updates hook after, with function', () => {
+      hooks.keep((item, context) => 'first')(hookAfter);
+      assert.deepEqual(hookAfter.result, { first: 'Jane' });
+    });
+
+    it('updates when called internally on server, with function', () => {
+      hookAfter.params.provider = '';
+      hooks.keep((item, context) => 'first')(hookAfter);
+      assert.deepEqual(hookAfter.result, { first: 'Jane' });
+    });
+
+    it('does not throw if field is missing, with function', () => {
+      const hook = {
+        type: 'before',
+        method: 'create',
+        params: { provider: 'rest' },
+        data: { first: 'John', last: 'Doe' } };
+      hooks.keep((item, context) => 'last', 'xx')(hook);
+      assert.deepEqual(hook.data, { last: 'Doe' });
+    });
+
+    it('keeps undefined values, with function', () => {
+      const hook = {
+        type: 'before',
+        method: 'create',
+        params: { provider: 'rest' },
+        data: { first: undefined, last: 'Doe' } };
+      hooks.keep((item, context) => 'first')(hook);
+      assert.deepEqual(hook.data, { first: undefined });
+    });
+
+    it('keeps null values, with function', () => {
+      const hook = {
+        type: 'before',
+        method: 'create',
+        params: { provider: 'rest' },
+        data: { first: null, last: 'Doe' } };
+      hooks.keep((item, context) => 'first')(hook);
+      assert.deepEqual(hook.data, { first: null });
+    });
+
+    it('keeps false values, with function', () => {
+      const hook = {
+        type: 'before',
+        method: 'create',
+        params: { provider: 'rest' },
+        data: { first: false, last: 'Doe' } };
+      hooks.keep((item, context) => 'first')(hook);
+      assert.deepEqual(hook.data, { first: false });
+    });
+
+    it('keeps 0 values, with function', () => {
+      const hook = {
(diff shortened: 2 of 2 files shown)


eddyystop:
I'm just thinking out loud here.

It can be agrued that every param for every hook should optionally be a function to allow dynamic values. Is a generic solution possible?

Perhaps dynamic params can be passed as getter props on an object?  Do you want to look in to that?

beeplin:
Yes it is a common need for many of the hooks to accept dynamic inputs. But for now I cannot think of a universal way to do this. One of the major problems is that different hooks need different inputs for the passed-in function. For example, `keep` uses `getItems` and `replaceItems` inside, so it is reasonable to accept a function taking `item` and `context` as params; however, say, `disallow` has nothing to do with `getItems` and so it just need a function with one `context` param. These two situations lead to quite different code inside the hook bodies, so I am not sure if there is a generic way to write them.

eddyystop:
I was thinking of something like
```js
const obj = {
  context: null,
  param1: () => this.context.foo + 5, // be as dynamic as you want
  param2: () => 6,
};

function myHook(a, b) {
  return context => {
    context.data.a = a; // 5
    context.data.b = b; // 6
    return context;
  };
};

all: [
  context => { obj.context = context },
  myhook(obj.param1, obj.param2)
];
```

I've only used getters a few times so I'm not sure if this code even works as is.

eddyystop:
My example doesn't work as in `myhook(obj.param1, obj.param2)` the `obj.param*` are evaluated when myHook is called, not when they are used in the inner func.

I still want to think about this for a while.

eddyystop:
Perhaps every param in the outer func of every hook should be converted to an optional func?

eddyystop:
Consider this
```js
// A basic sync identity function
function identity(store) {
  const prop = function() {
    if (arguments.length) store = arguments[0];
    return store;
  };

  prop.toJSON = () => store;

  return prop;
}

const param = identity();

let context = {};
const hooks = [
  context => { param(1); },
  myHook(param),
  context => { param(2); },
  myHook(param),
];

function myHook(val) {
  return () => { console.log(val()); };
}

hooks.forEach(func => {
  context = func(context) || context;
});
```

beeplin:
wow, what a wonderful design for the identity function~ like a magic...

beeplin:
> Hook param funcs would all have a signature of (context, item, info) => resolved_value. context, item, info are all objects. info is for misc properties in case they are needed in some cases.
item and info are undefined if they are not relavent for that hook param.
The resolved value may be a base type, object or function.
Hook param funcs may be async.

I am struggling with (context, item, ...) and (item, context, ...). Yes `context` is universally available and `item` is not, but in some cases `item` is relevant but `context` is not. Previously we always write something like `alterItems(item => ...)` which does not need to use `context` at all. If the params order is fixed to `(context, item, info)`, we have to write `alterItems((context, item) => ...)` or `alterItems((_, item) => ...)`, which is not friendly to eslint since `context` is an unused paramter.

I doubt that an object syntax might be a way to avoid this param order issue: `alterItems(({item}) => ...)`, `debug(({context, state, xxx, yyy}) => ...)`

I admit it looks ....even more ugly:((((

> Converting multiple params to optional func is a pain and it doesn't work well with TypeScript.

For this part, does that mean changing all/most of the current official hooks of feathers into this kind of hooks accepting the 'standard' format of param function? Or, even more, encourage users to use this param function signature as a convention to write their own hooks?

Do you mean that, for example, we should not have two types (overloads) of `keep` hook, one as `keep(...fieldNames)` and the other `keep((context, item) => ...)`, but have always only one `keep` with the function param? Even when we need `keep(...fieldNames)`, we need to use `keep(() => fieldsNames)` ?

I miss coffee-script so much, which neatly allows `keep => fieldNames`

eddyystop:
I should have been clearer. My comment with the sample code obsoletes my previous comment which mentioned (context, items). In fact, just ignore that previous comment.

Something like `context => { param(1); },` allows you to set the identity value to anything you want while having access to context and anything in it. There is no need to call a user provided function with (context, items, info) because `context => { param(1); },` has access to all that. Basically, instead of a function returning a value, the code updates the identity value.

So we would still have only one keep hook. We could call it with either `keep('first', 'last')` or with
```js
all: [
  context => { context. ... ? param(['first', 'last']) : param(...) },
  keep(param),
]
```

The code in keep would have to be changed to handle function (identity) parameters, as well as an identity value which is an array.

beeplin:
Thanks for the clarification. Yeah now I know where the identity function comes from. A wonderful idea, indeed~!

My only concern is that the `param` should not be *too global*. Maybe we could find a way to generate a *local* `param` for each hook-chain -- people must initialize `param` at the beginning of a hook chain in order to use it in the later hooks in the chain. So, basically for each service, we might be able to implicitly initialize 12 `param`s: before/after * create/patch/update/remove/find/get

Is this the right way to go?

eddyystop:
There is no limit to the number of fields you can have e.g. `keepFieldNames = indentity()` for `keep(keepFieldNames`; `lcFieldNames = identity()` for `lowerCase(lcFieldNames`. Just define the names you need for dynamic params you require.

There however is a major concern. Its possible for 2 calls on the same service to be running simultaneouly. We need to keep the identity functions separate for each call.

If 2 calls are performing the following, we cannot be sure `param` will be set properly for each hook chain.
```js
all: [
  context => { context. ... ? param(['first', 'last']) : param(...) },
  keep(param),
]
```
We need to think about this. The most obvious place to store the identity is in `context` but then I'm not sure how we refer to it in say `keep(xxx)`.

beeplin:
> If 2 calls are performing the following, we cannot be sure param will be set properly for each hook chain.

```js
all: [
  context => { context. ... ? param(['first', 'last']) : param(...) }, // 1
  keep(param),  // 2
]
```
I think node.js runs in a single tread, so as long as the function 1 is not async, it is guaranteed that 2 above runs immediately after 1 runs, and we don't need to worry about some other parallel calls changing the `param` between the two function calls.

if the function 1 is async, maybe it will also work if we write in this way:

```js
all: [
  async context => {
    param(context. ...  ? await someAsyncFunc(['first', 'last']) : await anotherAsynFunc(...context...)）
    keep(params())(context)
  }
]
```

I didn't make any test for this and am not sure if this would really work. But this makes me realize maybe we don't need to use the identity function at all because the two functions runs immediately one after the other. For sync functions, just use: 

```js
all: [
   context => { 
     const fieldsToKeep = context =>...
     keep(fieldsToKeep)(context)
  }
]
```

and for async functions: 

```js
all: [
   async context => { 
     const fieldsToKeep = async context => await someAsyncFunc(...) 
     keep(fieldsToKeep)(context)
  }
]

**Comment by beeplin** [comment]
But the solution above (and the identity function solution) cannot deal with the cases when we need to use both `context` and `item` to determine which fields to keep. For example, we have some data like this:

```js
console.log(await service.find())
// [{id: 1, a: ..., b: ..., fieldsToKeep: ['a']}, 
//  {id: 2, a: ..., b: ..., fieldsToKeep: ['a', 'b']}]
```

In my initial proposal we can do this:

```js
find: [ keep((item, context) => item.fieldsToKeep) ]
```

which returns:

```js
console.log(await service.find())
// [{id: 1, a: ...}, 
//  {id: 2, a: ..., b: ...]
```

Note that different items could return different fields according to its own `fieldsToKeep`

But in the solution mentioned above, `keep` always accept a fixed `fieldsToKeep` argument (which is dynamically generated according to `context`), and there is no way to let different items keep different fields.

eddyystop:
Your solution is much simpler.
```js
all: [
   async context => { 
     const fieldsToKeep = await someAsyncFunc(context) 
     keep(fieldsToKeep)(context)
  }
]
```

I cannot think of any common hook that acts differently for each record in item. How common is this? What is your use case?

beeplin:
I think the need to process each item differently is quite common. In my case, each user can choose what kind of there personal information is public (i.e. can be sent to other users when other users run app.service('users').find({....}). That's where the example above (keep with fieldsToKeep) comes from. At the beginning I was hoping to be able to write like this: 

```js
find: [ keep((item => ['name'].concat(item.fieldsToKeep)) ]
```

But never mind. Given the current `alterItems` and `keep`, we can always write in this way: 

```js
find: [
  alterItems((item, context) => {
    const targetField = context.type === 'before' ? 'data' : 'result'
    const fakeContext = {...context, {[targetField]: item}}  // fake a context as if it has only one item
    keep(item.fieldsToKeep)(fakeContext) // mutate the item using `keep`
    return pluckedItem = fakeContext[targetField] // let `alterItems` to replace the item with the new one
  })
]
```

Of coz in this case it is also easy to write my own keep/pluck logic instead of faking a context and using `keep`. But generally the given hooks in this repo is always much better established and tested, which is why I prefer using `keep` and `alterItems`.

eddyystop:
I'm not perturbed by the idea of building fake context objects. All the tests do it.

Does this mean we can close this issue by adding a section named 'Calling Hooks With Dynamic Options' to the common hooks docs, using your 2 examples?

beeplin:
sure

eddyystop:
@beeplin I have a question to ask that should be done over email. You want to, ping me at johnsz9999@gmail.com.

PR #335, opened 2018-01-17
Update makeCallingParams hook to use getByDot/setByDot

Mattchewone:
Resolves https://github.com/feathers-plus/feathers-hooks-common/issues/334


eddyystop:
Thanks for this.

Issue #337, opened 2018-01-19
StashBefore should skip the id check when a query is provided


Protoss78:
https://github.com/feathers-plus/feathers-hooks-common/blob/da263315e28fedb32c0cd0f11fff1a56a6bd635e/lib/services/stash-before.js#L11

The id validation happens before checking for the $disableStashBefore flag inside the query object. This makes it impossible to execute a clear all like functionality (eg. remove(null, {query:{}}) ), when a stashBefore hook is used in parallel. It should at least check for the existence of a query object and in parallel.

Protoss78:
Fixed in version 4.2.3.

PR #338, opened 2018-01-19
Fixed issue #337 and included relevant test - ignore missing id when …

Protoss78:
…query is used.

### Summary

PR fixess issue #337 
StashBefore now does not throw an error when the id is missing but a query parameter is present. 
Test script was updated with a scenario where all documents are removed, by passing no id and an empty query object.

### DIFF
--- a/lib/services/stash-before.js
+++ b/lib/services/stash-before.js
@@ -8,7 +8,7 @@ module.exports = function (prop) {
   return context => {
     checkContext(context, 'before', ['get', 'update', 'patch', 'remove'], 'stashBefore');
 
-    if (context.id === null || context.id === undefined) {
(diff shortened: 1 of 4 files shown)


eddyystop:
Do you think it would be cleaner to move
```js
if (context.id === null || context.id === undefined) {
  throw new errors.BadRequest('Id is required. (stashBefore)');
}
```
to after the folllowing if clause?

Could you bump the patch semver to 4.2.3.

And thanks for this catch!

Protoss78:
Added the version bump and moved the if statement. Definitely cleaner after the $disableStashBefore validation.

eddyystop:
Thanks for this!

Issue #345, opened 2018-01-31
Using softDelete and stashBefore results in infinite loop


leedongwei:
### Steps to reproduce
1. On the `before.all` hook, put in `softDelete` and `stashBefore`
2. Run a PATCH query.
3. Infinite loop happens

### Expected behavior
A preliminary GET request for data by 'before' hooks should be saved into the same location (e.g. context.params.before) to reduce the need for multiple GET from different hooks. 

### Actual behavior
1. User makes a PATCH request
1. User PATCH reaches softDelete hook
    1. softDelete makes GET request, add $disableSoftDelete flag
    1. softDelete GET reaches softDelete hook, [skipped due to flag](https://github.com/feathers-plus/feathers-hooks-common/blob/master/lib/services/soft-delete.js#L14)
    1. softDelete GET reaches stashBefore hook
        1. stashBefore makes GET request, add $disableStashBefore flag
          1. stashBefore GET reaches softDelete hook
              1. softDelete makes GET request... no flags to stop it, so infinite loop

### System configuration

**Module versions** (especially the part that's not working):
feathers-hooks-common v4.3.0

eddyystop:
Did you find a way around this?

leedongwei:
I was pressed for time so I just copied the source and hacked both hooks to fit my use-case. 

Is there a reason to delete the [$disableStashBefore](https://github.com/feathers-plus/feathers-hooks-common/blob/master/lib/services/stash-before.js#L12) and [$disableSoftDelete](https://github.com/feathers-plus/feathers-hooks-common/blob/master/lib/services/soft-delete.js#L15) flags? Keeping the flags in the params.query would solve the infinite loop issue.

I'd want to reduce the number of GET queries made by the hooks by stashing/checking context.params.before, but that would be quite an opinionated change. Not sure if you're up for that.

eddyystop:
It might be better to have a new softDelete which does its own stashing.

If you go that route, please review the existing softDelete issues. There are some interactions between softDelete for `users` and feathers-authentication.

bertho-zero:
I had the same issue, it's fixed by adding `params.query.$disableStashBefore = true;` after:
https://github.com/feathers-plus/feathers-hooks-common/blob/e67eb8ab413209caecc0fd95fd27021b16205436/lib/services/soft-delete.js#L64

We could add an argument to softDelete and stashBefore to add an optional query. Because as soon as we combine them the query of the get must be completed.

[...]

PR #346, opened 2018-02-01
Opinionated fix for #345

leedongwei:
### Summary

This pull requests fixes the following problems
1. Issue #345 
2. Multiple GET requests made by hooks for the same document

I introduced a new field, `context.params.before`, to stash the GET response and have the hooks check for it before proceeding further. I don't know how this will affect other hooks in the ecosystem, but I'd like to hear your take on it.


eddyystop:
Thanks for this.

I'll have to make time to consider the ramifications/

Issue #361, opened 2018-02-17
`mongoKeys` can also used in `before` `update/patch/remove`


beeplin:
Glad to see the new `mongoKeys` hook. That is what I always need since I transfered from `mongoose` to vanilla `mongodb`.

I will try using it to see if it can be improved. But at the first glance I can figure out that this hook can also be used before `update/patch/remove`. These three methods can also have `context.params.query` when `context.id === null`, and they do the exactly same thing as `find` does.

eddyystop:
Good point. Thank you.

PR #362, opened 2018-02-17
Allow `mongoKeys` for update/patch/remove

beeplin:
see #361


eddyystop:
Good point. Thank you.

Will be published in 4.7.1.

eddyystop:
Docs updated.

@beeplin, the guide now has a section on making hook params dynamic. Thanks for driving that.

PR #369, opened 2018-03-09
Proposal for #365: Prevent update changes

### LINKED ISSUE #365: preventChanges in update
There should be something like preventChanges for the update method as suggested [here](https://github.com/feathers-plus/feathers-hooks-common/issues/145#issuecomment-318769210).

hatsuo:
Resolves #365

### Summary

This is a proposal for the preventUpdateChanges hook discussed in #365.

The hook cannot be merged with the existent `preventChanges` hook, as the function signature is different (we need a `getFunc` parameter for the `update` method, which is not needed for `patch`).

If no `getFunc` is passed explicitely, the default function will be used which simply passes the hook context's `id` and `params.query` fields to the service call and can handle multi record changes.

I also didn't implement preventChanges's  `shouldThrow` parameter as this would cause trouble with multi record changes. This is the case, because if `shouldThrow` was set to `false` and only some of the requested records would be allowed to change, the hook would need to ensure that really only these records would be updated.

**NOTE**: I added [lodash.isequal](https://www.npmjs.com/package/lodash.isequal) as a dependency for the equality check as it already can handle object and array comparison.

### Other Information

This PR includes tests for the hook with the following coverage result:

| File                           |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
| -- | -- | -- | -- | -- | -- |
| prevent-update-changes.js    |      100 |      100 |      100 |      100 |                   |


eddyystop:
Thanks for the PR. I reviewed the code.

This hook is restricted to the `update` method. That method requires that an `id` be provided, and it will only update one row. Feathers otherwise throws `BadRequest: Not replacing multiple records. Did you mean `patch`?`.

Your code is based on `the notion that `update` can update multiple records, which is incorrect.

Once again, I question providing such a hook rather than requiring people to use the `patch` method. An alternative approach may be a hook which converts an `update` to a `patch` call.

hatsuo:
@eddyystop thanks I didn't know that, I changed the code accordingly. :)

I also added the `ifThrow` parameter from the `preventChanges` hook, since there is no issue with it anymore.

> Once again, I question providing such a hook rather than requiring people to use the patch method. An alternative approach may be a hook which converts an update to a patch call while removing any props for fields which may not be changed.

I see it the same way, I would also prefer to convert the requests into patches, but with that approach's given limitations I don't think that it is suitable for the majority, as it could induce misleading behavior and would eliminate some basic features of the update method.
But I'm open to being convinced otherwise. (hope this is correct englisch) :)

eddyystop:
Your English is excellent.

What limitations do you see if the hook converted the update call to a patch? The hook chain is created during `app` configuration, so the `update` hooks would be run even for `patch`.

That leaves common and custom hooks which act differently for `patch` than for `update`. A strong notice in the docs should be enough to warn people of that.

hatsuo:
Sorry, I don't know how to express this in a better way so I'll try to describe it with an example:

<details>
<summary>
Example
</summary>

Let's say we had an entry like this:
``` json
{
  "id": "1-2-3",
  "firstName": "John",
  "lastName": "Doe",
  "gender": "male",
  "settings": [ "..." ]
}
```

When we apply an `update` like the following...
``` json
{
  "firstName": "Bill",
  "lastName": "Gates"
}
```

... than we expect the result to look like something like this:
``` json
{
  "id": "1-2-3",
  "firstName": "Bill",
  "lastName": "Gates"
}
```
`id` stays as it is because it is a special field, but all other fields which miss in the `update` get omitted from the result.

But if we did a `patch` instead of an `update` with exactly the same request data we would get this:
``` json
{
  "id": "1-2-3",
  "firstName": "Bill",
  "lastName": "Gates",
  "gender": "male",
  "settings": [ "..." ]
}
```
Unlike the `update` method every field which was not described in the `patch` stays as it is and doesn't get omitted.

</details>
<br />

We have no way to unset fields or to set them to `undefined` if we only have `patch` but not `update`.

So the first problem would be that if we changed the `update` method to internally do a `patch` instead we wouldn't be able to unset fields anymore if we wanted to.

The second problem is shown in the example above.
When we do an `update` and don't mention a field in our request we expect this field to be definitely undefined in our result, even if it was set to a value before the `update`.
But `patch` works differently. It only modifies exactly those fields which were specified in the request and let's all other fields untouched. So if we changed the request into a patch internally we would change the behavior how these unspecified fields are treated.
And even if `patch` had a mechanism to unset fields we would nevertheless need to make a service call to fix this issue because we would need to find out which fields were omitted in the request and thus would explicitely need to be unset.

eddyystop:
"We have no way to unset fields or to set them to undefined if we only have patch but not update."
You can `service.patch(id, { name: undefined })`. I'm not sure if all the adapters consistently set the value to `undefined` or if they delete it.

I understand your points. I admit I was thinking more about adapters with schemas than MongoDB.

hatsuo:
Oh well I didn't know that.
I am mainly using ReST and I don't think that JSON is capable of doing something like this as it doesn't have an `undefined`, but I think on the serverside it should work fine.

So this still keeps the problem with schemaless adapters.

PR #371, opened 2018-03-10
alterItems now works with an async callback function

### LINKED ISSUE #367: Async functions in alterItems don't work
### Steps to reproduce
I ran the examples from https://feathers-plus.github.io/v1/feathers-hooks-common/#alteritems

You can find my code and output here:
https://gist.github.com/sean-nicholas/c0d73e386743a2b20f690aa473d88134

### Expected behavior
alterItems should work with async functions as described in the documentation and in https://github.com/feathers-plus/feathers-hooks-common/issues/307

### Actual behavior
Promises are not resolved as you can see in the gist output file.

### System configuration

**Module versions** (especially the part that's not working): 4.8.0

**NodeJS version**: 8.1.4

**Operating System**: macOS

sean-nicholas:
### Summary

alterItems did not work with a async callback function. See https://github.com/feathers-plus/feathers-hooks-common/issues/367
This PR fixes #367 

Documentation at: https://github.com/feathers-plus/docs/pull/19

### DIFF
--- a/lib/services/alter-items.js
+++ b/lib/services/alter-items.js
@@ -5,7 +5,7 @@ const replaceItems = require('./replace-items');
 
 module.exports = function (func) {
   if (!func) {
-    func = () => {};
+    func = () => { };
   }
 
   if (typeof func !== 'function') {
@@ -16,9 +16,14 @@ module.exports = function (func) {
     let items = getItems(context);
     const isArray = Array.isArray(items);
 
-    (isArray ? items : [items]).forEach(
-      (item, index) => {
-        const result = func(item, context);
+    const promises = (isArray ? items : [items]).map(
+      async (item, index) => {
+        let result = func(item, context);
+
+        if (result && typeof result.then === 'function') {
+          result = await result
+        }
+
         if (typeof result === 'object' && result !== null) {
           if (isArray) {
             items[index] = result;
@@ -29,7 +34,8 @@ module.exports = function (func) {
       }
     );
 
-    replaceItems(context, items);
-    return context;
+    return Promise.all(promises)
+      .then(() => replaceItems(context, items))
+      .then(() => context)
   };
 };
--- a/lib/services/alter-items.js
+++ b/lib/services/alter-items.js
@@ -1,4 +1,3 @@
-
 const errors = require('@feathersjs/errors');
 const getItems = require('./get-items');
 const replaceItems = require('./replace-items');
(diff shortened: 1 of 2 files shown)


eddyystop:
Feathers is supported on NodeJS v6 and v8. v6 does not support `async/await`.

The actual hooks cannot use `async/await` anywhere.

Its preferable the tests use `async/await` as little as possible. Where that is awkward, the tests can be placed in folder `tests-async` and called from `tests` using a stub like `fast-join-stub.test.js`.

sean-nicholas:
Ah okay. I will implement the changes asap :)

sean-nicholas:
Hey, I updated the code and the tests and removed every async / await

eddyystop:
Thank you.

eddyystop:
In v4.12.0. Docs updated.

bertho-zero:
This breaks my workflow with synchronous alterItems, they all turn into promises, it's a breaking change.

I think if there is no asynchronous then we should not return a promise.

sean-nicholas:
Hmm, I think it would be cleaner to change the docs. Because different return values based on the callback might be confusing:

* Callback func returns **not** a promise: **Don't** return a promise
* Callback func returns a promise: Return a promise
* Callback func returns a promise for some records: Return a promise

@eddyystop What do you think?

PS: How do you use alterItems? If you use it as normal hook it should simply work. Do you call it yourself in a custom hook?

eddyystop:
I don't think sync-type calls should return a Promise.

Maybe we need only remove the following as a non-Promise value is considered to be a resolved promise.

```js
        // If the result is not a promise, wrap it in one
        if (!resultPromise || typeof resultPromise.then !== 'function') {
          resultPromise = Promise.resolve(resultPromise);
        }
```
Can you modify the code and confirm this works @bertho-zero ?

sean-nicholas:
@eddyystop Don't get what you're trying to say 😄
What about this solution: https://gist.github.com/sean-nicholas/e93068b08239eac55b7140e31fb4c9c4

Be careful. I hacked it together. It does not use the hook context and therefore does not work with feathers. I just ran it in a quokka env to test it and therefore simplified the in- & output.

It now behaves as I outlined in my other comment:

- Callback func returns not a promise: Don't return a promise
- Callback func returns a promise: Return a promise
- Callback func returns a promise for some records: Return a promise

bertho-zero:
`hasPromises` is the most viable solution in my opinion. (I did not test the code)

eddyystop:
What you suggest should work. It was used at least in some old common hook code; maybe its still used.

I would suggest calculating `hasPromises` within the `map` loop.

I can fix the hook Wed or Thurs, else I can look at a PR.

[As for what I was trying to say, and I admit its a quibble, if you have `myPromise.then(a => console.log(a))` there is no guarentee that this `then` runs immediately after `myPromise` is resolved. A `then` or `catch` from another promise could run first.]

eddyystop:
@sean-nicholas, please check https://github.com/feathers-plus/feathers-hooks-common/pull/409 works with your code.

sean-nicholas:
#409 works for me 👍

PR #375, opened 2018-04-11
Fix deletion of array items

RubyRubenstahl:
### Problem
The `delete` operator is used to remove properties from objects and items from arrays. When `delete` is used on an array, the item is removed, but the array is not re-indexed;

```JS
let a=[1,2,3,4];
delete a[1];

// a now equals [1,,3,4]
```
If deleteByDot is used on an array that subsequently stored through a feathers service, the resulting array in the database (using the mongoose driver, at least) results in the "deleted" item being replaced with `null` rather than being completely removed. 

### Proposed Solution
The pull request tests to see if the item to delete is an array. If it is, `splice` is to remove the array item. Otherwise, `delete` is used. 

### Potential Problems
Because this results in different output, it could potentially break code that relies on the current behavior.

### DIFF
--- a/lib/common/delete-by-dot.js
+++ b/lib/common/delete-by-dot.js
@@ -16,5 +16,10 @@ module.exports = function (obj, path) {
     }
   }
 
-  delete obj[parts[nonLeafLen]];
+  if(Array.isArray(obj)){
+    obj.splice(parts[nonLeafLen],1);
+  }
+  else{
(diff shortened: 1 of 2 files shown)


eddyystop:
Thanks for this PR.

I'm leary of starting to add special cases to the `byDot` utilities, because this could become neverending. You did the work and added a test, so how can I say no?

PR #381, opened 2018-04-25
set-slug corrected for feathers buzzard

m4n0v31:
Use params.route[slug] instead of params[slug]

https://github.com/feathers-plus/feathers-hooks-common/issues/380


eddyystop:
I appreciate this.

Would you mind fixing the tests? These fail:

  services setSlug
    ignore feathers-socketio & feathers-rest clients
      1) ignore feathers-rest
    handles raw HTTP clients
      2) copies slug to query
    handles field name
      3) copies slug to query
    handles field name with dot notation
      4) copies slug to query

PR #390, opened 2018-05-05
use concat when possible

bertho-zero:


bertho-zero:
On travis:
```
error clone-deep@3.0.1: The engine "node" is incompatible with this module. Expected version ">=8".
```

eddyystop:
I'm not sure using concat everywhere produces better performance. I also think `xxx ? yyy : [yyy]` is faster to understand.

PR #395, opened 2018-05-29
Update all dependencies and badges

daffl:
This pull request

- Removes code climate since it is no longer reporting
- Update the Readme badges to the current repository location


eddyystop:
Thank you.

An obscure issue was reported with ajv 6.0.x. We'll see about 6.5.0.

PR #396, opened 2018-05-30
Enable Semistandard and update code style

daffl:
Adding Semistandard to the other repos always helped avoid bugs and encouraged contributors to follow a common code style. It was removed from the standard test script but it was passing without problems when I tried running it so this PR proposes to enable Semistandard again and includes the automated fixes.


eddyystop:
Thank you.

PR #399, opened 2018-06-04
keep hook conserve null result

bertho-zero:
Closes https://github.com/feathers-plus/feathers-hooks-common/issues/394.


eddyystop:
Could you change this to
```js
if (typeof item !=== 'object' || item === null) return item;

and at the bottom

return newItem;
```

PR #400, opened 2018-06-04
closes #368

### LINKED ISSUE #368: `discard` fails on null properties
### Steps to reproduce

Using the `discard` hook, if a parent property is `null`, the underlying `getByDot` function fails due to [a typeof check](https://github.com/feathers-plus/feathers-hooks-common/blob/master/lib/common/get-by-dot.js#L8) improperly passing.

Sample failure: https://gist.github.com/alexisabril/74d33d2b30331849e0e9ed73fda8cb72

### Expected behavior

If a property on the path happens to be null, `getByDot` should return `undefined` as opposed to trying to access the property.

### Actual behavior

As `getByDot` currently checks `typeof obj1 === 'object`, a `null` will cause a truthy result. The ternary response in this instance would be `null[part]`, which results in an error.

`deleteByDot` covers this scenario: https://github.com/feathers-plus/feathers-hooks-common/blob/master/lib/common/delete-by-dot.js#L14

A solution is to modify the condition to match `deleteByDot`:

```
if (typeof obj !== 'object' || obj === null) ...
```

bertho-zero:


eddyystop:
Also https://github.com/feathers-plus/feathers-hooks-common/pull/403

PR #405, opened 2018-06-13
remove hook removal explained

rayfoss:
Save someone a couple minutes. It's one of the most popular hooks.


bertho-zero:
This is the discard hook, the delete hook does not exist.

PR #406, opened 2018-06-16
Fix prevent changes and discard hooks

SteffenLanger:
### Summary
Enable the hooks discard and preventChanges to delete properties containing a dot.

**Problem**
This is about nested objects in PATCH requests. A PATCH request for MongoDB can either replace an entire object through a request like this:
`{
  "contactPerson": {
    "lastName": "Doe",
    "firstName" "John"
  }
}`
or it can update an individual property like this:
`{
  "contactPerson.firstName" "Jane"
  }
}`
In the latter case, only the property `firstName` is will be updated, while `lastName` will stay intact.

The MongoDB documentation explains this here: https://docs.mongodb.com/manual/reference/operator/update/set/#up._S_set ([feathers-mongoose](https://github.com/feathersjs-ecosystem/feathers-mongoose) implicitly wraps the object from a PATCH request with a {$set : ...} object).

Currently, however, both `discard('contactPerson.firstName')` and `preventChanges(false, 'contactPerson.firstName')` only catch the first PATCH object and ignore the second one.

**Expectation**
The discard hook would remove both property formats from the PATCH request to prevent updating e. g. passwords, usernames etc.
I would bet that quite a few systems are currently subject to an attack because their owners assume that both discard and preventChanges remove the second PATCH format as well.

**Solution**
Both hooks check whether there is a property containing the dot on the object itself. If so, delete that property.
The previous behavior remains unchanged: The dot notation will **also** traverse the object tree. So, both context.data.contactPerson.firstName and context.data['contactPerson.firstName'] will be caught.

- [ ] Are there any open issues that are related to this? No.
- [ ] Is this PR dependent on PRs in other repos? No.

### Other Information

If there's anything else that's important and relevant to your pull
request, mention that information here. This could include
benchmarks, or other information.

Your PR will be reviewed by a core team member and they will work with you to get your changes merged in a timely manner. If merged your PR will automatically be added to the changelog in the next release.

If your changes involve documentation updates please mention that and link the appropriate PR in [feathers-docs](https://github.com/feathersjs/feathers-docs).

Thanks for contributing to Feathers! :heart:

### DIFF
--- a/lib/common/_remove.js
+++ b/lib/common/_remove.js
@@ -5,6 +5,9 @@ const _transformItems = require('./_transform-items');
 module.exports = function (items /* modified */, fieldNames) {
   _transformItems(items, fieldNames, (item, fieldName, value) => {
     if (value !== undefined) {
+      // Delete fields like item['contactPerson.name']. This is a typical patch format for MongoDB.
+      if(item[fieldName]) delete item[fieldName];
+      // Delete fields like item.contactPerson.name
       deleteByDot(item, fieldName);
     }
   });
--- a/lib/services/prevent-changes.js
+++ b/lib/services/prevent-changes.js
@@ -18,15 +18,15 @@ module.exports = function (...fieldNames) {
     const data = context.data;
 
     fieldNames.forEach(name => {
-      // Delete data['contactPerson.name']
-      if(data[name]) {
-          delete data[name];
-      }
-      // Delete data.contactPerson.name
-
       if (existsByDot(data, name)) {
         if (ifThrow) throw new errors.BadRequest(`Field ${name} may not be patched. (preventChanges)`);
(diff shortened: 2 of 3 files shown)


eddyystop:
Could you fix the tests?

  1) services preventChanges
       checks props
         throw if props found:
     AssertionError: expected [Function] to throw an error
      at Context.it (tests/services/prevent-changes.test.js:49:19)
  2) services preventChanges
       throws if first param is "true"
         throw if props found:
     AssertionError: expected [Function] to throw an error

SteffenLanger:
Thank you for your feedback!

I'll have a look at the tests this week.

eddyystop:
I like these changes. Thanks! Published as 4.14.3 .

PR #409, opened 2018-06-19
restore alterItems for works with sync methods

bertho-zero:


lib/services/alter-items.js
```diff
@@ -15,29 +15,33 @@ module.exports = function (func) {
...
-        return resultPromise.then(result => {
-          if (typeof result === 'object' && result !== null) {
-            if (isArray) {
-              items[index] = result;
-            } else {
-              items = result;
-            }
-          }
+    if (hasPromises) {
+      return Promise.all(results.map(v => Promise.resolve(v)))
```
sean-nicholas: No need to map all results to promises. Promise.all can handle non promises.

eddyystop:
Thanks much. Published as v4.12.1

PR #410, opened 2018-06-20
Update soft-delete.js

### LINKED ISSUE #257: SoftDelete catches all errors
I have a question, why catches the softDelete hook all errors, when it gets an item?
https://github.com/feathersjs/feathers-hooks-common/blob/62f554dd0a6a15509248caae6d472e7801858451/src/services/soft-delete.js#L77
If there is an error in the hook, its not throwing that error, instead 404 not found.

### LINKED ISSUE #389: softDelete does not trace the errors that occur in the get call
`softDelete` does not trace the errors that occur in the get hooks.

If a real error other than NotFound occurs in the get there is currently no way to know it, it would be necessary to "swallow" the error rather than replace it, I use [verror](https://github.com/joyent/node-verror) for that in all my projects.

https://github.com/feathers-plus/feathers-hooks-common/blob/1942e7c0087c3761db31558cba4e8fad52947ad6/lib/services/soft-delete.js#L66-L77

bertho-zero:
Closes #257 
Closes #389

### DIFF
--- a/lib/services/soft-delete.js
+++ b/lib/services/soft-delete.js
@@ -68,12 +68,9 @@ module.exports = function (field) {
           delete params.query.$disableSoftDelete;
 
           if (data[deleteField]) {
-            throw new errors.NotFound('Item has been soft deleted.');
+            throw new errors.NotFound('Item not found.');
           }
           return data;
-        })
-        .catch(() => {
-          throw new errors.NotFound('Item not found.');
         });
     }
   };


eddyystop:
Wouldn't you still need a catch to avoid the dreaded `UnhandledPromiseRejectionWarning`?

eddyystop:
Thank you.

I'll wait on publishing this is you plan to sumbit more PRs.

PR #411, opened 2018-06-26
Added callingParams, callingParamsDefault hooks. Replaced makeCallingParams

eddyystop:
See docs for new hooks `callingParams` and `callingParamsdefault`.

`makeCallingParams`  is now a stub to `callingParams`. It is fully compatible with the previous 'makeCallingParams` hook.

`callingParams` should now be used instead of `makeCallingParams` as the former
- has more features
- common hooks will use it internally


eddyystop:
Published as 4.13.0

PR #412, opened 2018-06-26
Update calling-params.js

bertho-zero:


eddyystop:
Those are not params but properties of the object parameter, so no default need be defined.

Thanks for looking at the code, and it should be useful for softDelete.

PR #413, opened 2018-06-29
Hook for signing in via "Basic" authentication in a "before" hook

bitflower:
# Summary

Hey Eddy, as discussed on [Slack](https://feathersjs.slack.com/archives/C0HJE6A65/p1530216141000160?thread_ts=1530126436.000082&cid=C0HJE6A65) here is the hook. 

I have the following questions:
- Is the way of actually signing in correct? (via the `/authentication` service?)
- I think we should make the props `username` and `password` flexible according to the `@feathers/configuration` JSON file. What do you think?
- Does this fit the codebase (ES5, ES6, ..?)

# Usage

```javascript
import { hooks } from '@feathersjs/authentication';
import { basicAuth } from './basic-auth';

// Combined call to basic auth hook and authenticate
const customAuth: any[] = [basicAuth(), hooks.authenticate('jwt')];

// For example to protect write actions:
export default {
    before: {
        all: [],
        find: [],
        get: [],
        create: [...customAuth],
        update: [...customAuth],
        patch: [...customAuth],
        remove: [...customAuth]
    },

    after: {
        all: [],
        find: [],
        get: [],
        create: [],
        update: [],
        patch: [],
        remove: []
    },

    error: {
        all: [],
        find: [],
        get: [],
        create: [],
        update: [],
        patch: [],
        remove: []
    }
};

```

I :heart: Feathers ;-)

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -210,7 +210,7 @@ export function getItems(context: HookContext): any; // any[] | any | undefined;
  * Check which transport provided the service call.
  * {@link https://feathers-plus.github.io/v1/feathers-hooks-common/index.html#IsProvider}
  */
-export function isProvider(...transports: TransportName[]): PredicateFn;
+export function isProvider(...transports: TransportName[]): boolean;
 
 /**
  * Keep certain fields in the record(s), deleting the rest.


eddyystop:
Replies to questions:
1. [Is the way of actually signing in correct? (via the /authentication service?)]
It would be best to ask @daffl.
2. The consistent way would be to pass the prop names as parameters to the basicAuth call, likely within an options object.
3. The code has to run on Node 8 as is, and transpile via Babel ES5.

Could you add tests? You'd also have to
1. Add the module to lib/services/index.js (in alphabetical order please)
2. Add the module to tests/services/exposed.test.js (in alphabetical order please)

eddyystop:
Have you considered what can be done for logging off Basic Auth?

Here's my response on SO on how to do it https://stackoverflow.com/questions/16360293/how-to-log-out-from-basicauth-express/17172222#17172222

bitflower:
1. Sorry for the delay, no I haven't thought about logging out. Isn't there a timeout once there is no activity anymore with the user?

You mean a hook that would be added as `pre`/`post` on the logout route? Good idea! Let me write it.... Will take another few days.

2. Yes, sure, but how do I access the settings from `default.json`? Are they inside `context.app.something`? Gotta check....

3. Travis failed with Node 6 though? Seems like an old dependency... Checking.

4. Tests: Will do when everything else is cool.

bitflower:
@daffl Any thoughts on 1. ? Would be appreciated!

eddyystop:
"how do I access the settings from default.json?" via `app.get(propname)`.

bitflower:
I've saved the conversation to a PDF because I'll have to close this PR for now.

PR #414, opened 2018-07-02
[security issue] fix populate injecting provider into source schema argument

antarasi:
### Summary

Populate hook is modifying original `options` argument in place causing provider field being injected into all subsequent populate calls. 

If the first populate call has been done with `provider = undefined` (for instance by restrictToOwner hook), populate inject this provider into original `options`. The second call from rest provider is performed with `provider = undefined` because that value was injected into `options` in previous popolate call. 

Such behavior is a severe security issue making possible to fetch related objects bypassing owner check.

More detailed description is given here:
[https://bitbucket.org/AntaraSI/feathers-populate-test](https://bitbucket.org/AntaraSI/feathers-populate-test)

### DIFF
--- a/lib/services/populate.js
+++ b/lib/services/populate.js
@@ -59,8 +59,7 @@ module.exports = function (options, ...rest) {
             if ('provider' in schema) {
               return schema;
             } else {
-              schema.provider = provider;
-              return schema;
+              return Object.assign({}, schema, {provider});
             }
           });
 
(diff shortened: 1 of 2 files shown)


eddyystop:
You spent a long time in tracking this issue down. Thank you very much for that.

Published as 4.14.1

antarasi:
True... Almost two weeks of tracking random "No auth token" errors. Thanks

PR #420, opened 2018-07-27
Added `keepInArray` hook to keep fields from a nested array

dekelev:
### Problem
The `keep` hook does a great job for its intended purpose, but it won't work for nested arrays. for example, finding a record with one-to-many or many-to-many ORM relations.

### Solution
This new `keepInArray` hook works like `keep`, but it accepts 2 arguments. the first is a dot notation string type field and the second is an array of field names to keep from it.

### Usage

Data structures:
```js
{ 
  users: [
    {
      name: 'John',
      secret: 'doNotKeep',
      dept: 'Accounting',
      address: {
        city: 'New York',
        secret: 'doNotKeep'
      }
    }
  ],
  owner: 'Jane Doe'
}

{
  account: {
    users: [
      {
        name: 'John',
        secret: 'doNotKeep',
        dept: 'Accounting',
        address: {
          city: 'New York',
          secret: 'doNotKeep'
        }
      }
    ],
    owner: 'Jane Doe'
  }
}
```

Using hook:
``` js
const { keepInArray } = require('feathers-hooks-common');
    
module.exports = { after: {
  create: keepInArray('users', ['name', 'dept', 'address.city']),
  find: keepInArray('account.users', ['name', 'dept', 'address.city']),
} };
```

Data structures after using hook:
```js
{ 
  users: [
    {
      name: 'John',
      dept: 'Accounting',
      address: {
        city: 'New York'
      }
    }
  ],
  owner: 'Jane Doe'
}

{
  account: {
    users: [
      {
        name: 'John',
        dept: 'Accounting',
        address: {
          city: 'New York'
        }
      }
    ],
    owner: 'Jane Doe'
  }
}
```

Added tests and [PR](https://github.com/feathers-plus/docs/pull/43) to the feathers-plus docs.


eddyystop:
First of all, great job figuring out how to change the docs, and following the repo standards for the hook code.

Dot notation and arrays have been a quandry. In the general case of `a.b.c.d.e` both `b` and `d` may be arrays. I did not want to start dealing with multiple arrays.

Am I correct in thinking your hook will throw on line 22 if `field` is not an array? Would it be helpful if an explicit error identifying the problem is thrown instead?

Your comments are welcome, and let me think on this also.

dekelev:
@eddyystop I think you right. we should throw error there, but since it can be both for data or result, I'm not sure about the correct type of error to thrown. BadRequest? GeneralError?

What do you suggests?

eddyystop:
I think the more clear a name is, the more likely a hook is to be found by users. Would you mind changing the name to keepInArray? I'll merge thereafter.

dekelev:
@eddyystop sure, name changed to `keepInArray` (also in the docs PR)

eddyystop:
Published as 4.16.0

PR #427, opened 2018-08-14
Enhanced cache with MongoDB/Mongoose support

eddyystop:
option.makeCacheKey allows bson keys to be converted to strings.


eddyystop:
Fizes https://github.com/feathers-plus/feathers-hooks-common/issues/424

PR #431, opened 2018-08-23
Added `keepQuery` support for keeping props with dot in their name

dekelev:
With ORM libraries (e.g. feathers-objection), the client sometimes need to query using a full field name (e.g. `'user.id=1'`) or filter by relation field (e.g. `'dept.name=Acct'`) and it will populate the query object in the server like this: `{ query: { 'user.id': 1 } }`

`keepQuery` hook currently only performs `getByDot` & `setByDot` using the `_pluck` module and these example queries will fail, since the `keepQuery` hook will always remove them.

I would like to be able to use `keepQuery('user.id', 'dept.name')` and would expect that the `_pluck` module will try simple `get & set` before resorting to the `getByDot & setByDot` hooks.

### DIFF
--- a/lib/common/_pluck.js
+++ b/lib/common/_pluck.js
@@ -21,9 +21,9 @@ function _pluckItem (item, fieldNames) {
 
   fieldNames.forEach(fieldName => {
     const valueNotByDot = item[fieldName];
-    const value = valueNotByDot || getByDot(item, fieldName);
+    const value = typeof valueNotByDot !== 'undefined' ? valueNotByDot : getByDot(item, fieldName);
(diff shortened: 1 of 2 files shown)


eddyystop:
I don't think this works for existing falsey values.

eddyystop:
Thisa may be a quibble, but I would prefer using something like `fieldName in item` rather than something like `typeof valueNotByDot !== 'undefined'`, as that's the test used by `existsByDot`.

dekelev:
I've changed validation to use `in`

eddyystop:
Thanks!

Published as 4.16.3

PR #444, opened 2018-09-17
fix: isProvider returns a boolean

bitflower:
Hi Eddy,

as discussed on Slack, I think the `isProvider` method is returning a `boolean`.

The usage of it should therefore be changed in the docs as well I guess to something like:
```
iff(
        (context) => isProvider('external'), ...
```

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -210,7 +210,7 @@ export function getItems(context: HookContext): any; // any[] | any | undefined;
  * Check which transport provided the service call.
  * {@link https://feathers-plus.github.io/v1/feathers-hooks-common/index.html#IsProvider}
  */
-export function isProvider(...transports: TransportName[]): SyncPredicateFn;
+export function isProvider(...transports: TransportName[]): SyncContextFunction<boolean>;
 
 /**
  * Keep certain fields in the record(s), deleting the rest.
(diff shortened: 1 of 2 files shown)


types/index.d.ts
```diff
@@ -210,7 +210,7 @@ export function getItems(context: HookContext): any; // any[] | any | undefined;
  * Check which transport provided the service call.
  * {@link https://feathers-plus.github.io/v1/feathers-hooks-common/index.html#IsProvider}
  */
-export function isProvider(...transports: TransportName[]): PredicateFn;
+export function isProvider(...transports: TransportName[]): boolean;
```
j2L4e: this is not correct, it should return "SyncPredicateFn", would you also please fix the test?

j2L4e:
test still needs fixing, change the expected type to `SyncContextFunction<boolean>`

j2L4e:
don't worry. I was referring to `tests.ts`, you can inspect the error by clicking on the red cross on the right hand side of your commit.
Right now the test still expects `PredicateFn` which is wrong, we'd need it to expect `SyncContextFunction<boolean>` (which `SyncPredicateFn` is basically an alias for)

bitflower:
Fixed tests.ts. Travis looks good?

j2L4e:
Great, thanks :)
@eddyystop lgtm

eddyystop:
Published as v4.17.5

PR #450, opened 2018-09-26
Correctly bind hooks and predicate functions used in conditionals.

### LINKED ISSUE #258: iff doesn't bind hook functions to service instance
when hooks are called by the iff function `this` is not bound to the service the hook was run on.

### Steps to reproduce

```
app.service('myService').hooks({
  before: {
    all: {
      function (hook) {
        console.log(this === hook.app.service('myService'));
      },
      iff(() => true, function (hook) {
        console.log(this === hook.app.service('myService'));
      }),
    }
  }
});
```

prints
```
true
false
```

j2L4e:
fix #258 

better late than never!

### DIFF
--- a/lib/common/iff.js
+++ b/lib/common/iff.js
@@ -1,10 +1,8 @@
 
 module.exports = function (_iffElse) {
   return function (predicate, ...trueHooks) {
-    const that = this;
-
     function iffWithoutElse (context) {
-      return _iffElse(predicate, [].concat(...trueHooks), null).call(that, context);
+      return _iffElse(predicate, [].concat(...trueHooks), null).call(this, context);
     }
     iffWithoutElse.else = (...falseHooks) => _iffElse(predicate, [].concat(...trueHooks), [].concat(...falseHooks));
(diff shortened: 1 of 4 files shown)


eddyystop:
This is an error. However "this" does not work in other conditions hooks such as iff().else(). I've some time trying to fix them. Do you want to show your prowess?

j2L4e:
Seems to me like it works. I'll have a closer look this weekend

eddyystop:
There were various problems getting `this` to work the conditional hooks. I don't see the fix in `iff`, which is wrapper over `iffElse`, solving them all. So please check carefully.

From a higher level view though, people now use `context.app` and `context.service`. I don't want to encourage the use of the basically deprecated `this`, though its nice to support "legacy" hooks inside conditionals.

j2L4e:
Closing this PR as it's not worth the needed effort. If you need to support a legacy hook inside conditionals, consider writing a wrapper hook that binds `context.service` to your legacy hook.

PR #460, opened 2018-11-01
Allow $populate: object, don’t skip the hook

marshallswain:
Changes behavior of the fgraphql hook with regards to `context.params.$populate`.  Instead of skipping requests where `$populate` exists, this skips requests where `$populate` is either an array or a string, thus allowing `$populate: { }`

```js
$populate = {
  email: 1,
  posts: {}
}
```


eddyystop:
context.params.$populate is a boolean used by the populate, fastJoin and fgraphql hooks.

PR #463, opened 2018-11-14
remove dtslint workaround in favor of npx

### LINKED ISSUE #461: dtslint-build folder in installed package?
This is a bit strange, but I was trying to install `feathers-hooks-common` into a project using `pnpm` and `rush.io` and it would constantly choke on the installation of this library with `EEXIST: file already exists` error and the following (truncated) output:
```
/common/temp/pnpm-store/2/registry.npmjs.org/feathers-hooks-common/4.17.14/node_modules/feathers-hooks-common/dtslint-build/node_modules/typescript/lib/pt-br/diagnosticMessages.generated.json' -> 
/common/temp/node_modules/.registry.npmjs.org/feathers-hooks-common/4.17.14/node_modules/_tmp_27828_842f75df51d0959c837427c958974785/dtslint-build/node_modules/typescript/lib/pt-br/diagnosticMessages.generated.json'
```

After a little bit of investigation, it looks like there's a `dtslint-build` folder that is being inserted even though it doesn't exist within the actual Github repo. Because this `dtslint-build` folder has it's own `node_modules` directory in it, it seems like that is what `pnpm` is choking on. This is from a fresh `npm init && npm i -d feathers-hooks-common`:

screen shot 2018-11-02 at 3 51 24 pm

I assume this is happening because some sort of cleanup function from after tests/builds run in your CI/CD is being missed and the artifacts are being committed to the final build?

j2L4e:
work in progress.

fixes #461, but requires npx (npm@^5.2.0) to be present, so it ~could~ will break type linting for people without npx installed.

The solution was taken from https://github.com/ReactiveX/rxjs/pull/4088 as it was the only workaround I could find and I trust the rxjs guys' expertise. If someone can come up with a better workaround please step forward.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,4 +31,3 @@ node_modules
 
 dist/
 .idea/
-dtslint-build/
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "release:major": "npm version major && npm publish",
     "changelog": "github_changelog_generator && git add CHANGELOG.md && git commit -am \"Updating changelog\"",
     "lint": "semistandard --fix",
-    "lint:types": "mkdirp dtslint-build && cd dtslint-build && echo '{\"devDependencies\":{\"dtslint\":\"^0.3.0\"},\"scripts\":{\"dtslint\":\"dtslint ../types\"}}' > package.json && npm install && npm run dtslint",
+    "lint:types": "npx dtslint types",
     "mocha": "mocha --opts mocha.opts",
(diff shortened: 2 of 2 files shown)


j2L4e:
Linting runs fine on travis but it doesn't seem to be working for me locally with TS@next. I'll investigate that.

**Do not merge.**

j2L4e:
I've tried it several times today and it seems to run fine locally. CI works too. 
@eddyystop this can be merged :)

PR #470, opened 2018-11-23
required() of value false throws fieldname is null

antarasi:
When trying to require boolean fields, value `false` should not thow an error.
Haven't found any related issues.
It's independent PR.


eddyystop:
Thank you!

PR #477, opened 2018-12-02
add typings for dialablePhoneNumber hook, add libphonenumber-js as a dependency

j2L4e:


j2L4e:
https://github.com/feathers-plus/feathers-hooks-common/issues/474

PR #479, opened 2018-12-04
Proposal: Use Lodash for common utility functions

daffl:
[Lodash](https://lodash.com/) contains a lot of the utility functions that are implemented again as common hooks utilities. I am proposing to use the Lodash methods instead for the following reasons:

- Lodash is probably one of the most battle tested and micro-optimized npm modules and covers many edge cases we may only run into with weird bugs
- When I tried to use the existing methods (`get-by-dot` instead of `_.get`) they behaved slightly different than what I was expecting (e.g. it is not possible to escape `.` and objects with dot string names have to be handled as a special case - `_.get` does all of that already)
- Lodashs `_.get` and `_.set` are faster since it has a mechanism for caching object paths
- Less code (and typings) to maintain

All tests for the proposed changes here are still passing (except for the deprecated deleting undefined values for `set-by-dot`).

### DIFF
--- a/lib/common/exists-by-dot.js
+++ b/lib/common/exists-by-dot.js
@@ -1,17 +1,5 @@
+const has = require('lodash/has');
 
 module.exports = function (obj, path) {
-  const parts = path.split('.');
-  const nonLeafLen = parts.length - 1;
-
-  for (let i = 0; i < nonLeafLen; i++) {
-    let part = parts[i];
-
-    if (!(part in obj)) { return false; }
-
-    obj = obj[part];
-
-    if (typeof obj !== 'object' || obj === null) { return false; }
-  }
-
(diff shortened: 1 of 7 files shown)


eddyystop:
lodash is certainly solid. I didn't use lodash because
1. The hooks can and are being used client. lodash has a larger footprint even though its broken up now.
2. I know at least some people are using the utilities directly. I'm not sure the set utility, for example, is 100% compatible with lodash's.

I'll look at deeper into this.

daffl:
Yep, 1) could be a little bit of a drawback. I'm importing the dependencies specifically so it should only pull in what is needed. For 2) I would say it only supports what is tested. Probably should be released as a breaking change anyway.

PR #496, opened 2019-01-04
move `@types/graphql` package from `devDependencies` into `dependencies`

### LINKED ISSUE #487: graphql dependency
Correct me, but should not graphql module be a dependency of feathers-hooks-common instead of devDependency?!

Because we now get build errors in our typescript project. It tells us the following:

```
node_modules/feathers-hooks-common/types/index.d.ts(5,60): error TS2307: Cannot find module 'graphql'
```

j2L4e:
fixes #487


eddyystop:
Published as v4.19.8

PR #506, opened 2019-01-23
Fix promise-to-callback hook preventing other hook import.

rybaczewa:
### Summary

Current version of promise-to-callback hook breaks importing in TypeScript. For projects that run in environment where neighter `process` nor `setImmediate` exist, you cant do:
`import { paramsForServer } from 'feathers-hooks-common';`
(in my case, angular 6 application)

Code compiles but `setImmediate is not defined` error is thrown. Even without using this particular hook.

### DIFF
--- a/lib/services/promise-to-callback.js
+++ b/lib/services/promise-to-callback.js
@@ -1,11 +1,11 @@
 
 /* globals setImmediate:1 */
 
-const asap = process && typeof process.nextTick === 'function'
-  ? process.nextTick : setImmediate;
-
 module.exports = function (promise) {
   console.log('**Deprecated** The promiseToCallback hook will be removed next FeathersJS version.');
+
+  const asap = process && typeof process.nextTick === 'function'
+    ? process.nextTick : setImmediate;
 
   return cb => {
     promise.then(


eddyystop:
@j2L4e you use Angular. Could you comment please?

j2L4e:
Don't think this is angular specific, it's rather client specific:
https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate

`setImmediate` seems to be largely unsupported in non-node environments

j2L4e:
`babel-polyfills` includes this automatically. TS doesnt.

add `import 'core-js/web/immediate`;` to your `polyfills.ts`

j2L4e:
To address the PR: It would fix importing it, but you'd still be unable to use the hook. Not a proper fix.

I think it should just use `fn => setTimeout(fn, 0)` if `setImmediate` isn't available.

rybaczewa:
@j2L4e Considering this hook is being deprecated, I went for easiest and least invasive change. Yours should too. No matter which one we go for, I think this code still should be moved inside of the function. I don't see why it should run even without importing this specific hook.

eddyystop:
@rybaczewa moving the code inside the hook resolves your issues?

NickBolles:
@eddyystop from just looking at this, it stops the error from being thrown when importing the file, but it will still throw an error when actually running it.

eddyystop:
The method is deprecated as Node's new utility should be used.

So there should be no new uses of the method and any existing uses will still work.

NickBolles:
@rybaczewa correct me if I'm wrong, but it looks to me like that error will throw whenever anything from feathers-hooks-commons is imported, making this an issue for all of feathers-hooks-commons, not just this  hook. 

Moving the offending reference to setImmediate into the hook call hides this until the hook is called. If the hook is called you will still get the message. 

The real fix is to do something like this, I'm not sure this is 100% right, but it should be close

```
  const supportsNextTick =  (process && typeof process.nextTick === 'function')
  const supportsSetImmediate = (typeof setImmediate === 'function')
  const timeoutFallback = (cb, delay, args) => setTimeout(() => cb(args),delay) // lamda function to immitate setImmediate/process.nextTick, use () => cb(args) to preserve "this"? I think?
  const asap = supportsNextTick ? process.nextTick : (supportsSetImmediate ? setImmediate :  timeoutFallback);
```

rybaczewa:
@NickBolles that's correct, I just moved it inside of the hook. Mainly to:
1) unblock import of other hooks
2) don't change any behaviour of any of the hooks
3) don't spend to much on hook that's being deprecated.

So either this or full fix simillar to yours. That's just maintainer decision, made the pull request to showcase the problem and propose possible solution.

eddyystop:
Are you both agreed this PR is good enough during the limited life of this hook?

NickBolles:
Right. I think this is fine if this hook is deprecated.

PR #509, opened 2019-02-14
Update iff predicates typings to be either sync or async

NickBolles:
### Summary
The iff* hooks were typed to only accept synchronous predicate functions
This PR adds
1. Typescript type definitions for Asynchronous (in addition to the previous synchronous), Predicate functions
2. Typescript type tests (in types/tests.ts) to test for Async Predicates
3. 2 more tests for Iffelse to get it to 100% code coverage in normal (non-typescript) tests
No
no

### Other Information

There are a few small updates to docs to happen, there are more after my next PR

#### iffElse
arguments - `{Function} predicate` -> `{Boolean | Promise | Function} predicate`

#### iffElse
arguments - `{Function | Boolean} predicate` -> `{Boolean | Promise | Function} predicate`

### DIFF
--- a/tests/services/iffelse.test.js
+++ b/tests/services/iffelse.test.js
@@ -65,6 +65,28 @@ describe('services iffElse', () => {
     hookFcnAsyncCalls = 0;
   });
 
+  describe('runs single hook', () => {
+    it('when true', () => {
+      return hooks.iffElse(true, hookFcnSync, hookFcnAsync)(hook)
+        .then(hook => {
+          assert.deepEqual(hook, hookAfter);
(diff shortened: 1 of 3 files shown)


eddyystop:
The predicate can already be a function which returns a Promise, hence the function can be asynchronous, even an `async` function. Its not clear what feature this PR adds.

NickBolles:
@eddyystop I think you miss-read the title. This is for the typescript types. They were previously restricted to synchronous predicate functions, which is incorrect.

I added the typescript types for both Sync or Async Predicate functions, as well as Typescript type tests for IFF, and a normal test to make test coverage 100%. (I'll update the OP with this info)

PR #510, opened 2019-02-14
WIP: Simplify and standardize predicates

NickBolles:
### Summary

This is abstracting predicates to be run in 1 spot, in a very clean way. Hopefully makes predicates easier to maintain (not that they were hard to maintain, but eh)
Yes #509 

### Other Information
Basically this makes it so that predicates are called with runPredicate, and can be sync or async and uses the shared predicate in `skip-remianing-hooks` and `isNot`

it also makes it really easy to use predicates and predicates on arrays (multi-predicates like some, every etc)

Documentation updates:

#### IsNot
- Update predicate parameter to be async or sync

#### skip-remaining-hooks
- Update predicate parameter to be async or sync

#### some,every
- Update predicate parameter to be async or sync


NickBolles:
My main question with this is, is this change okay?

Basically all of the hooks that returned sync results when sync predicates were passed in will no longer be sync. They will return promises. 

I updated the skipRemainingHooks tests to account for promises and not check for the Sync function returning a sync result. But I wanted to wait to get feedback before going through the rest. (there's about 30 other tests that are currently failing)

Is this a breaking change? is there actually a use case for running a hook in sync mode? IMHO it just seems to complicate things. This PR will bring together all of the predicate running into a few very simple lines of code that are shared across all hooks. It will also allow some more powerful patterns, such as predicates inside of predicates and predicate functions that return promises (which could also be predicates). How useful those will actually be? I haven't put enough thought into. But hey. It's possible.

eddyystop:
I would think many more sync hooks are used rather than async ones. It may not matter for performance since, as you say, promises created with a value are immediately resolved, plus the hooks are run a promise chain anyway.

However people do run common hooks from within custom hooks and this would break that code. Therefore I will tag this for when we next make breaking changes to the repo.

Please submit a localized PR if you have a sync/async issue with a particular hook.

PR #515, opened 2019-04-25
Ensure `existsByDot` and `getByDot` honour properties with dots in keys

anthonygood:
Resolves: https://github.com/feathers-plus/feathers-hooks-common/issues/514

Origin of this change is the [`keep()`](https://feathers-plus.github.io/v1/feathers-hooks-common/#keep) hook failing for object keys containing dots (like `{ 'foo.bar': true }`).


marshallswain:
Thank you!

PR #541, opened 2019-11-14
Use Lodash for utility functions

daffl:
[Lodash](https://lodash.com/) is a widely used, highly tested and performance optimized library. This PR removes the custom utility functions and uses their Lodash equivalent.


marshallswain:
:shipit:

PR #543, opened 2019-11-20
Safely delete properties using lodash/omit

daffl:
This pull request removes the `delete-by-dot` and related helpers and instead removes properties safely by using [lodash _.omit](https://lodash.com/docs/4.17.15#omit). As with the move to other Lodash utilities in #541 it has been thoroughly tested, performance optimized and comes with predictable, well documented behaviour.

Most importantly, this change does no longer mutate existing objects which can cause many hard to debug issues and unexpected behaviour.


marshallswain:
:shipit:

PR #555, opened 2019-11-27
modify stashBefore hook, move $disableStashBefore from query to params

marshallswain:
This moves the custom param `$disableStashBefore` to actually be part of the params instead of messing with the `params.query` object.

~I'm going to redo the tests, so please do not merge.~

Update: I was able to update the existing tests.  This is ready.


lib/services/stash-before.js
```diff
@@ -24,12 +23,10 @@ module.exports = function (prop) {
     };
 
     params.query = params.query || {};
-    params.query.$disableStashBefore = true;
+    params.$disableStashBefore = true;
 
     return context.service.get(context.id, params)
       .then(data => {
-        delete params.query.$disableStashBefore;
```
marshallswain: @daffl what are your thoughts about leaving the param alone?  We needed to delete this when it was part of `params.query`, but I'm leaning towards leaving this here.  Any downsides you can think of?

daffl: Yes, we can leave it. Also, I would just use `params.disableStashBefore` as the name, I don't think it needs the `$`.

lib/services/stash-before.js
```diff
@@ -24,12 +23,10 @@ module.exports = function (prop) {
     };
 
     params.query = params.query || {};
-    params.query.$disableStashBefore = true;
+    params.disableStashBefore = true;
 
     return context.service.get(context.id, params)
```
daffl: Here is a good example where a small change (not mutating existing objects) can make a big difference when it comes to avoiding subtle bugs. Instead of modifying `params` we could probably just do:

```js
return context.service.get(context.id, {
  ...params,
  query: params.query || {},
  disableStashBefore: true
})
```

marshallswain: I've updated the PR.

PR #565, opened 2020-01-22
Change check for external provider as per daffl in disallow.js by 104c1b1

jalbersdorfer:
Change check for external provider as per daffl in disallow.js by 104c1b1

### Summary

fixes a 'undefined' to 'boolean' cast flaw in the same way it was already done before for the `disallow` hook.


daffl:
Thank you!

PR #603, opened 2020-08-18
fix(): No patch method

eduardoRoth:
When using softDelete in `before` `application hooks` for some services that don't have patch method (for example feathers-authentication-management service) it will fail.

### Summary

- [X] Tell us about the problem your pull request is solving.
When using softDelete as an application hook in combination with `feathers-authentication-management` the `logout` method will fail because softDelete hook would try to patch but `authentication` service doesn't have a patch method.

- [X] Are there any open issues that are related to this?
Not that I found

- [X] Is this PR dependent on PRs in other repos?
No

### Other Information

If there's anything else that's important and relevant to your pull
request, mention that information here. This could include
benchmarks, or other information.

Your PR will be reviewed by a core team member and they will work with you to get your changes merged in a timely manner. If merged your PR will automatically be added to the changelog in the next release.

If your changes involve documentation updates please mention that and link the appropriate PR in [feathers-docs](https://github.com/feathersjs/feathers-docs).

Thanks for contributing to Feathers! :heart:

### DIFF
--- a/lib/services/soft-delete.js
+++ b/lib/services/soft-delete.js
@@ -34,9 +34,11 @@ module.exports = ({
 
     if (method === 'remove') {
       const data = await getValue(removeData, context);
-      const result = await service.patch(context.id, data, params);
 
-      context.result = result;
+      try {
+        const result = await service.patch(context.id, data, params);
+        context.result = result;
+      } catch (err) {}
     }
 
     return context;


daffl:
Thank you for the pull request. I'm not sure I understand why you would use softDelete in this case. I don't believe it will work at all if there is no patch method.

eduardoRoth:
@daffl let me elaborate more:

I'm using the package `feathers-authentication-management` which creates a new service in path `authentication` and/or `authManagement` which handles the required logic for signup, login, logout and more.

I'm using the `softDelete` hook with custom properties for all my services so, instead of adding it in each `.hooks` file, I added it to my application hooks.

This works fine for everything, until I try to logout. The logout method in `feathers-authentication-management` is a `.remove` one so it triggers the softDelete hook which changes the remove to patch. But as the authentication management service doesn't have a patch method, it throws an error.

This change is to avoid having to set my `softDelete` hook on each service so that it doesn't break the authentication management logic.

daffl:
That makes sense but your pull request will just silently swallow _any_ error that happens when removing an item. I recommend to instead run it conditionally in your `app.hooks`:

```js
iff(context => context.path !== 'authentication', softDelete())
```

eduardoRoth:
Weird, when I first tried using `iff` the `softDelete()` method wasn't called. Thanks, that'll fix it.

PR #605, opened 2020-09-04
Fix multi record remove cache

uulwake:
### Summary

- [ ] Tell us about the problem your pull request is solving.
Currently, if we want to remove multi-record in the cache it will throw error `TypeError: Cannot read property 'id' of undefined`. It is caused because we are trying to access `getItems`  in before remove hook which will return `null`. We can only do `getItems` in after remove hook. Therefore, we need to skip to clear cache in before remove hook and do that in after remove hook instead.

- [ ] Are there any open issues that are related to this?
No

- [ ] Is this PR dependent on PRs in other repos?
No

### DIFF
--- a/lib/services/cache.js
+++ b/lib/services/cache.js
@@ -16,7 +16,14 @@ module.exports = function (cacheMap, keyFieldName, options = {}) {
     const query = context.params.query || {};
 
     if (context.type === 'after') {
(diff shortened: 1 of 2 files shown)


daffl:
This is great, thank you! Will go out as a release shortly.

PR #607, opened 2020-09-08
fix(docs/cache): replace 'clone' in arguments table by 'makeCacheKey'

H3r3zy:
I just replace 2nd 'clone' option in arguments table of 'cache' hook by 'makeCacheKey'.
There was 2 options named 'clone'


daffl:
Thank you!

PR #623, opened 2021-04-24
Dove Support

forgot:
Updates `@feathersjs` dependencies to `^5.0.0-pre.3` to work with Dove
Not that I could find...
Nope!

This pull request primarily updates the type expectations to match Dove. Due to the removal of the `processHooks` function, the `combine` hook had to be reimplemented. It would obviously be better if this remained it's own branch and did not merge with `master`, but I'm not sure how to create a pull request that does that.

All tests pass with `npm test`


daffl:
This is also great, thank you! Probably needs the same `Paginate` change we discussed the other day and that's now published.

forgot:
I've bumped everything to `5.0.0-pre.4` and moved the `Paginated` import back. All tests are still passing.

forgot:
I've bumped everything to 5.0.0-pre.9, and all tests are still passing. I'm guessing all of these will need to be migrated to the new format with backwards compatibility before this is of any real use in Dove.

daffl:
I wonder what the best approach would be here. Should all hooks be updated to the new format and provide backwards compatible `before` and `after` wrappers or be forward compatible by checking for the `next` function like in https://github.com/feathersjs/feathers/blob/dove/packages/authentication-local/src/hooks/protect.ts#L16-L18

I'm also wondering if this might be a good time to look at things that can be improved using the new hooks or consolidated.  For example, many common hooks for string casing and array processing could just be wrappers for [standard Lodash](https://lodash.com/docs/4.17.15) and I think there is also some overlap with @DaddyWarbucks great [feathers-fletching](https://github.com/DaddyWarbucks/feathers-fletching).

fratzinger:
Could we have a pre-release of this?

daffl:
I believe the latest release should work. Does it not?

fratzinger:
I want to update my libraries ([feathers-casl](https://github.com/fratzinger/feathers-casl) and [feathers-utils](https://github.com/fratzinger/feathers-utils)) to feathers@5.

image

Screenshot shows type differences between `HookContext` without generics from feathers@5 that's passed to `checkContext` of `feathers-hooks-common`. When checking out the `forgot:Dove` branch, this error is gone.
(btw. sorry that it's in german)

fratzinger:
Closing in favor of #691. Thanks for your work!

PR #635, opened 2021-07-23
Update migrating.md

indatawetrust:
since the setField function has been moved into the feathers-authentication-hooks package, I have added this to the documentation


daffl:
Thank you!

PR #662, opened 2022-05-04
Update index.d.ts

indatawetrust:
### Summary

hi all, anchor link usage in code comments problem fixed. current usages do not work properly as all links in the document are in lowercase


fratzinger:
Thanks @indatawetrust for your work! We merged this with #663 which is a typescript refactor.

PR #682, opened 2022-07-26
refactor(set-field hooks): Handle 'as' field startsWith data and data type array

arnoldtkl:
To handle the use case where `multi:true`, `create` method, the data is in array format. E.g
Inside hook
```
before: {
  all: [],
  find: [],
  get: [],
  create: [setField({ from: "params.user.id", as: "data.user_id" })]
  update:[],
  patch: [],
  remove: []
}
```
When POST array of data, currently the merge result will become
```
[ 
  { key: data, key2: data },
  { sample: data, sample2: data },
  user_id: "sample_id"
]
```
Expected result:
```
[ 
  { key: data, key2: data, user_id: "sample_id" },
  { sample: data, sample2: data, user_id: "sample_id" }
]
```

### DIFF
--- a/src/hooks/set-field.ts
+++ b/src/hooks/set-field.ts
@@ -22,7 +22,7 @@ export function setField (
   }
 
   return context => {
-    const { params, app } = context;
+    const { data, params, app } = context;
 
     if (app.version < '4.0.0') {
       throw new Error('The \'setField\' hook only works with Feathers 4 and the latest database adapters');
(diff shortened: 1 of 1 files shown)


arnoldtkl:
Thanks for the quick reply! I agree with you, `setData` is more straightforward and declarative. Maybe we can make a note at the `setField` documentation, e.g when dealing in `multi:true` service , the result may not work as expected, we will recommend check on `setData`  ......

By the way, I will close this pull request. Because my code didn't cover nested arrays and not handling well in the `as` field format.

fratzinger:
Very good point to add it to the docs. Would you be up to add a PR for that? Any help is highly appreciated.

arnoldtkl:
Ok, sure. But, the setData not yet included in this library. I will write a note about handling `multi:true` may not work as expected.

PR #712, opened 2023-04-04
Pass additionalProperty name in nested objects to validateSchema error messages

ardcore:
This fixes a bug in `validateSchema` hook where error messages related to invalid additional properties in nested objects didn't contain the name of said erroneous property.


fratzinger:
Thanks for the PR, especially for the added test!

Released as `v7.0.3` 🎉

PR #718, opened 2023-05-05
new: next

FossPrime:
### Summary

Adds a hook named next. It attaches an iterator named next to paginated results to improve pagination ergonomics.

https://github.com/feathersjs-ecosystem/feathers-hooks-common/issues/717

### DIFF
--- a/src/hooks/next.ts
+++ b/src/hooks/next.ts
@@ -1,0 +1,20 @@
+import type { HookContext } from '@feathersjs/feathers';
+
+/**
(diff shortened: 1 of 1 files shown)


FossPrime:
Given the racy nature of Pagination and variables like Mongo's estimatedTotalCount...

@fratzinger Could we put this into an experimental labs export? Kinda like this, example, where DataTables API is in flux and being iterated over for years.

Screenshot_20230520-172355

PR #751, opened 2024-05-24
fix: handle actOn dispatch correctly in replace items

### LINKED ISSUE #750: update replaceItems util to support pagination on dispatch
replaceItems correctly handles pagination for [context.result](https://github.com/feathersjs-ecosystem/feathers-hooks-common/blob/1b9e4c58a9ae813d5abfee89d88d32a6ade3933a/src/utils/replace-items.ts#L17), but not for dispatch.

jd1378:
resolves #750


jd1378:
hi @fratzinger 
can we have this merged and released ?

fratzinger:
Released as `v8.1.3` 🎉 Thanks for the change. Sorry for the delay and thanks for the ping 👍
Would you be up to add tests for this?

jd1378:
Hi
thanks for the release
check #752 for tests
I tried it by reverting the change locally and the test fails, so it should be good

PR #752, opened 2024-06-10
chore: update tests for replaceItems with actOnDispatch

jd1378:
### Summary

Tests for #751

### DIFF
--- a/test/hooks/get-replace-items.test.ts
+++ b/test/hooks/get-replace-items.test.ts
@@ -1,5 +1,5 @@
 import { assert } from 'vitest';
-import { getItems, replaceItems } from '../../src';
+import { actOnDispatch, getItems, replaceItems } from '../../src';
 
 // Tests when context.params._actOn === 'dispatch' are in act-on.test.ts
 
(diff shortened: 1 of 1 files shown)


jd1378:
well I just copied the test block above it and modified it for dispatch, but It looks like an important test so it might be intentional (?)

fratzinger:
lol. merged it then. I want to work on a few things in this repo anyways. I don't like the implicit magic that happens around `replaceItems`. When to use `context.data`, when to use `context.result` depending on `context.type`? It does not work for around hooks at all? What do you want to achieve with `replaceItems` in a `before` hook if there already is `context.result` because you want to skip the adapter call. 

Too many questions and to much magic for my taste. I want to introduce a more explicit `replaceData` and `replaceResult` and maybe `replaceDispatch` or get that covered by `replaceResult(..., { useDispatch: true | false | 'ifExistent' })` or something. Just thinking out loud.

That said I'll eventually rework this stuff anyways.

jd1378:
hmm, I found the bug for replaceItems when I was using feathers-casl. it was using replaceItems but things were not getting replaced. so It took me some time to pin point it to the dispatch issue. and then I had to patch feathers-casl to replace both result and dispatch. I have done A LOT of modifications to feathers-casl to get it working the way I want it to, and it's still not what I expect.
I know it's not the place to discuss it, just wanted to mention where this little utility was having impact on my project.
(btw feathers-casl's most interesting feature for me was the `getChannelsWithReadAbility` function)

fratzinger:
Wrote you on discord. I would love to chat about this stuff. Not now but in the near future.

PR #753, opened 2024-07-17
Fix/neotraverse

fratzinger:
see this: https://github.com/puruvj/neotraverse

if you missed it, this is the drama behind the update of `traverse@0.6.8` -> `traverse@0.6.9`: https://x.com/passle_/status/1810805530706792930

see current graph of feathers-hooks-common:
Bildschirmfoto 2024-07-17 um 19 55 53
source: https://npmgraph.js.org/?q=feathers-hooks-common#zoom=h


fratzinger:
released as `v8.2.0`

image

marshallswain:
Much better.  :grin:

PuruVJ:
Thanks a lot for considering neotraver 😄! That graph looks svelte and lean!

fratzinger:
That was fast. I didn't think you'll see this. Thanks for your work! 💚

eXigentCoder:
Hey there, thanks for the great work!
I think there might be an issue with this though (but perhaps I just need to change something my side), it looks like it's importing the `ESM` version of traverse and the minified output contains the export keyword:

`... minifiedCode;export{m as Traverse,v as default};`

Should the import statements in `feathers-hooks-common` not be changed to this instead?:

`import traverse from 'neotraverse/legacy';` 

If not, I'm getting an error in my project when I run my Jest tests 
```SyntaxError: Unexpected token 'export'```
Coming from `node_modules/neotraverse/dist/min/index.js` which is in my project due to `feathers-hooks-common@8.2.0`

PuruVJ:
Ah that’s correct, it should be legacy if you’re still shipping commonJS
builds

fratzinger:
released as `v8.2.1` 🎉 @eXigentCoder please try it out and let me know.

PR #755, opened 2024-07-21
feat: neotraverse/legacy; v0.6.14

PuruVJ:
### Summary

- Bumps neotraverse to 0.6.14 minimum
- Use neotraverse/legacy as the build still relies on CommonJS. Issue caught in https://github.com/feathersjs-ecosystem/feathers-hooks-common/pull/753#issuecomment-2241460378

Closes #753

PR #756, opened 2024-07-21
feat: neotraverse/legacy; v0.6.14

fratzinger:
This is a follow up of #755 which changes the new 'neotraverse' dependency to 'neotraverse/legacy' for cjs modules.

#755 couldn't be merged because of a regression in node@22.5.0 (see https://github.com/npm/cli/issues/7657). 22.5.1 is out and fixes this but it's not picked up by gh action 'actions/setup-node' yet. So as a temporary fix I set v22.5.1 explicitly in the 'node-version' matrix. We'll need to revert it to '22.x' in a future PR.

Thanks @PuruVJ for the fix! Thanks @eXigentCoder for the report in #753. 

Closes #755

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -7,9 +7,9 @@ jobs:
       matrix:
         node-version: [18.x, 20.x, 22.x]
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


fratzinger:
released as `v8.2.1` 🎉

PR #766, opened 2025-03-25
fix iff hook parameters

bburns:
the docs for the iff hook seem to have the wrong parameters, copied from iffElse?


fratzinger:
Thanks!

